PR #9, opened 2015-07-20
Python2to3

VelizarVESSELINOV:


VelizarVESSELINOV:
http://stackoverflow.com/questions/11914472/stringio-in-python3

kinverarity1:
Thanks - I've made a slight modification to accommodate Python 2.7 systems that don't have an optimized cStringIO module in the standard library -- my change is in a new PR (#12) because I got confused about how github works. Can you see if that works on Python 3? 

If so I will merge #12  to master and update PyPI as suggested to ensure the library is available for Python 2 and 3.

Thanks for the help!

VelizarVESSELINOV:
I checked yesterday the branch Python2to3 and worked well.

Thanks.

PR #40, opened 2015-08-03
Possible fix for name problem

diverdude:
Im guessing this was just supposed to be mnemonic instead of name?


VelizarVESSELINOV:
I have a question if mnemonic is the preferred term, the function should be ``get_curve_mnemonic`` instead of ``get_curve_name``?

diverdude:
i agree

kinverarity1:
I have some improvements but will add later.

PR #61, opened 2015-08-09
Support reading version 3.0 of LAS file spec

kinverarity1:
Starting with some unit tests for the spec examples.

Still a long way off being supported.

### DIFF
--- a/lasio/las.py
+++ b/lasio/las.py
@@ -204,7 +204,7 @@ class LASFile(OrderedDictionary):
 
         # Validate version
         try:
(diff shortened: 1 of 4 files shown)


ahjulstad:
Just to let you know, I am working on some preliminary LAS3 reading support. [branch](https://github.com/ahjulstad/lasio/tree/limited-las3)

It will probably involve using structured dtypes for the numpy array. The one LAS3 example I have (added) is also abusing string fields for dates, not sure how to deal with that.

PR #63, opened 2015-08-09
LAS -> Excel converter script

kinverarity1:
See #8 for original issue.

Will add ``xlwt`` as a requirement, but it's fairly lightweight so I don't see as an issue.


kinverarity1:
Tests pass for everything except Python 3.2 -- apparently ``xlwt`` has problems with that.

But as it turns out ``xlwt`` is somewhat outdated anyway - I'll investigate using either [``openpyxl``](http://openpyxl.readthedocs.org/en/latest/) or [``xlsxwriter``](https://xlsxwriter.readthedocs.org/) instead.

PR #72, opened 2015-08-11
Retain commented-out lines in header

kinverarity1:
When reading LAS files there may be comment lines in a variety of positions in the header. Ideally `lasio` should retain these lines so that when the file is written back to disc they are placed in as close to their original position as possible.

An example is in the test suite ~Well section:

```
~WELL INFORMATION BLOCK
#MNEM.UNIT       DATA TYPE    INFORMATION
#---------    -------------   ------------------------------
 STRT.M        1670.000000:
 STOP.M        1660.000000:
 STEP.M            -0.1250:
  ## BACKWARDS
 NULL.           -999.2500:
 COMP.             COMPANY:   # ANY OIL COMPANY LTD.
 WELL.                WELL:   ANY ET AL OIL WELL #12
 FLD .               FIELD:   EDAM
# Edam field sounds like an interesting place...
 LOC .            LOCATION:   A9-16-49-20W3M
 PROV.            PROVINCE:   SASKATCHEWAN
 SRVC.     SERVICE COMPANY:   ANY LOGGING COMPANY AT ALL!!!!
# Which is strange, right?
 DATE.            LOG DATE:   25-DEC-1988
 UWI .      UNIQUE WELL ID:   100091604920W300
# This unique well ID does not exist... or does it?
# This is just to test the comment-reading/writing functionality.
~CURVE INFORMATION
#MNEM.UNIT      API CODE      CURVE DESCRIPTION
#---------    -------------   ------------------------------
 DEPT.M                      :  1  DEPTH
 DT  .US/M               :  2  SONIC TRANSIT TIME
```

I think the cleanest way to implement this is to actually read and add the lines as items in the `OrderedDictionary` for each section. This would vastly simplify writing them back out. However this ties in with plans I have been mulling over to subclass `OrderedDictionary` in order to provide a variety of things to do with the `LASFile` object (e.g. partly broken item/attribute access). It should tie in with those things.

### DIFF
--- a/lasio/test_examples/readwritecomments.las
+++ b/lasio/test_examples/readwritecomments.las
@@ -1,0 +1,40 @@
+~VERSION INFORMATION
+ VERS  .                  1.2:   CWLS LOG ASCII STANDARD -VERSION 1.2
+ WRAP  .                  NO:   ONE LINE PER DEPTH STEP
+~WELL INFORMATION BLOCK
+#MNEM.UNIT       DATA TYPE    INFORMATION
+#---------    -------------   ------------------------------
+ STRT.M        1670.000000:
+ STOP.M        1660.000000:
+ STEP.M            -0.1250:
+  ## BACKWARDS
+ NULL.           -999.2500:
+ COMP.             COMPANY:   # ANY OIL COMPANY LTD.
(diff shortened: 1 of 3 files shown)


kinverarity1:
This should be considered as part of a comprehensive improvement of the write method.

PR #73, opened 2015-08-11
automatically provide depth index in metres and feet (fixes #15)

### LINKED ISSUE #15: Automatically provide depth index in feet and metres
Would be handy.

kinverarity1:
see #15


VelizarVESSELINOV:
I'm sorry to be very picky, but the conversion factor between feet and meter is not 100% correct, it will be not accepted by any major oil company.

The official conversion factor is defined by Energistics (for Oil and Gas, but inherits all the basic SI and US standards) and it is ``0.3048`` and not ``3.28084`` which is approximation of ``1/0.3048=3.28083989501``. The data managers of the oil companies are very picky on this subject and this is part of the first question they ask which factors you are using to convert units, if the answer is not
standard shame on you.
http://www.energistics.org/asset-data-management/unit-of-measure-standard

Side note: it will be nice if you can build another package that manage all the conversions and you encapsulate with lasio for extra conversion not only depth, it will be also very useful for validation of units, ``F`` Farad is used too often in US for ``ft``, ``MV`` often is used instead ``mV``, etc.

VelizarVESSELINOV:
Another comment the spelling of ``ft`` (feet, FT, F, FEET...) or ``m`` (M, METERS, METER, metre, meter, м, метер in Cyrillic...) are very broad, today you are covering ``FT`` and some case like parameters ``F``. It will be nice to provide consistent 

Section 2 is talking about unit aliases, not sure if Energistics is providing official list of unit aliases (like we have for characters codecs):
http://w3.energistics.org/uom/Energistics_Unit_of_Measure_Usage_Guide_V1.pdf

It will be nice to be able to fix replace unit of LAS files without touching the rest of the file, for example replace "F" with 'ft' and keep the file as it is with all the comments and errors.

For this kind of LAS cleaning in the past I done a simple read/write line by line parser and changed with regular expression if 'F' was in unit section with 'ft' as an example.

Another warning: there are a lot of Energistics units with '.' and ' ' (like ``Pa.s`` or ``0.001 psi/ft``) in middle and should be not treated as LAS section separator in the middle of the unit.

To be honest we need the same for mnemonics, I found a lot of strange mnemonics with special characters like tab, quotes, comma, etc.

Energistics is working for new version of PWLS: http://w3.energistics.org/PWLS/pwls_20.htm that will be helpful to search for Gamma Ray type of mnemonic in LAS file.

kinverarity1:
> The official conversion factor is ... 0.3048 and not 3.28084

Fixed in 82e6999. That was an oversight of my part in not using a good source for the conversion factor.

> nice if you can build another package that manage all the conversions and you encapsulate with lasio for extra conversion not only depth, it will be also very useful for validation of units

Agreed, it would be an interesting project. The Energistics site has a lot of useful data to base it on.

> the spelling of ft (feet, FT, F, FEET...) or m (M, METERS, METER, metre, meter, м, метер in Cyrillic...) are very broad, today you are covering FT and some case like parameters F

I based it on the LAS 2 specification as a starting point: "If the index is depth, the units must be M (meters), F (feet) or FT (feet)". However your suggestion is a good one - I have opened issue #74 for it.

> It will be nice to be able to fix replace unit of LAS files without touching the rest of the file, for example replace "F" with 'ft' and keep the file as it is with all the comments and errors.

PR #72 will result in the write method retaining all commented lines, but it will not retain other errors in formatting.

> a simple read/write line by line parser and changed with regular expression if 'F' was in unit section with 'ft' as an example.

Excellent solution but I'll leave that for others to contribute for now.

> there are a lot of Energistics units with '.' and ' ' (like Pa.s or 0.001 psi/ft) in middle and should be not treated as LAS section separator in the middle of the unit.
> 
> To be honest we need the same for mnemonics, I found a lot of strange mnemonics with special characters like tab, quotes, comma, etc.

This is a big problem I've run into already. It is going to be difficult to write a reliable parsing regular expression, but it is something that will need to be tackled. I've raised issue #75 for it.

> helpful to search for Gamma Ray type of mnemonic in LAS file.

I will keep an eye on it -- that kind of use case is something I would very much like ``lasio`` to provide in the future.

VelizarVESSELINOV:
Thank you for all the comments/reactions/feedback.

The LAS 2.0 clarification about "F", "M", "FT" was very interesting, I understand better why some LAS files are conflicting with POSC/Energistics specifications. I hope CWLS and Energistics will work better together in the future to standardize these units spellings issues.

To be honest a lot of files don't follow these constraints and with LAS 3.0 you will need probably to be more agile on the reference unit spelling.

PR #106, opened 2015-09-20
Internal changes looking to LAS 3.0 support

kinverarity1:
Changing sections internally from OrderedDicts to lists, where some access as a map is implemented in a list subclass (SectionItems).

Passing basic tests but a lot still to clean up.


kinverarity1:
New aspects:

- Sections are now represented by a list subclass ``SectionItems``, which contains only ``HeaderItem``s (or in the future subclasses of that)
- Sections are kept in ``LASFile.sections`` dictionary - the ``LASFile.well`` etc dictionaries are properties that send you to ``LASFile.sections['Well']``

Issue #94, opened 2015-08-16
Create pandas.DataFrame attribute storing log data


kinverarity1:
This used to be in there but it made more sense to require only numpy.

However, if the user has pandas available then we should use it. Call the attribute ``df`` and use the reference curve as the index for the DataFrame.

kwinkunks:
Just wanted to +1 this. It would make a great optional dependency. 

I don't know that these will help, but to illustrate a use case, the notebooks [here](https://github.com/seg/tutorials/tree/master/1504_Seismic_petrophysics_1) and [here](https://github.com/seg/tutorials/tree/master/1506_Seismic_petrophysics_2) use pandas for some rock physics calculations.

kinverarity1:
Still to-do before this issue is closed:

- Add example to the documentation
- The only unit test at the moment tests that the reference curve from the LAS file is correctly used as the index in the DataFrame and Series, but I need to add tests for the presence of the curves and also a reverse-direction LAS file index -- although it will probably work it's not being tested.

kinverarity1:
Curve names also checked now - 0082159

PR #110, opened 2015-12-02
Add dataframe example to one of the notebooks

dagrha:
kinverarity you mentioned in issue #94 that an example of the pandas dataframe attribute was needed for the documentation.  Here I've added an example to one of your detailed usage notebooks.  Was this what you had in mind?


kinverarity1:
That's fantastic, thank you!

dagrha:
My pleasure, thank _you_ for creating and maintaining this great project.

I just put the dataframe example into one of your existing notebooks, but I'd be happy to spin it off into its own notebook if you'd prefer, or move the examples to a different notebook, or add more examples of dataframe manipulation, etc.

kinverarity1:
I think what you've put in there is excellent where it is as a basic 
example... But if you have a more extended example of typical things you do 
with a dataframe, absolutely it would be very welcome. Probably easiest as 
a separate notebook.

I'll probably reorganise that existing one sometime as I think we need a 
simpler introduction, less about the header items.

Also - I'm not at all familiar with the full functionality of pandas, so if 
you come across any small or large ways in which lasio could be extended or 
changed to accommodate working with data inside the dataframe, please 
please open issues for them.

Issue #107, opened 2015-11-03
LASFile object should be serializable with json.dumps()


knbknb:
I'd like to dump the lasfile into Redis (a popular nosql database), or save it to disk in the platform-independent JSON Format. 
At this time, this is not possible. 

<pre>
import redis
import json
r = redis.StrictRedis(host='localhost', port=6379, db=0)
#json_images = json.dumps(lasobj) # DOES NOT WORK
json_images = json.dumps(lasobj, default=lambda x: None)
r.set('lasobj', json_images)
    TypeError: array([ 134.5946,  134.62  ,  134.6454, ...,  745.6424,  745.6678,
        745.6932]) is not JSON serializable
</pre>

kinverarity1:
I'll fix this - it should work directly with ``json.dumps()``.

kinverarity1:
#111 provides a partial fix:

```python
json_images = json.dumps(lasobj, cls=lasio.JSONEncoder)
```

will work (a unit test for it has been added).

However the JSON representation only contains the data, not the header information.

I think a better solution would be to provide a ``LASFile.json`` attribute which covers all the information in the file. I'll leave this issue open for now.

kinverarity1:
I should align this with whatever format JustinGOSSES/wellio.js settles on.

kinverarity1:
This looks sensible: https://github.com/JSONWellLogFormat/JSONWellLogFormat

We should provide a `LASFile.to_json()` method which uses the serializer above.

And, why not, a `from_json()` class method.

dcslagel:
https://github.com/JustinGOSSES/wellio.js/pull/36 is an exploration and discussion PR about the json format for JustinGOSSES/wellio.js.  

It includes:
-  wellio_sample_2.0.json : a full wellio json file based on sample_v2.0.las
- wellio.schema.json : an initial sketch of what a Wellio son-schema might look like.  In the first commit, only the Version section is specified.

Is this useful with regards to Lasio's json roadmap?

DC

kinverarity1:
Absolutely - looks great! 

I don't know anything about JSON schemas, so haven't got much to directly contribute to the discussion, but I certainly would prefer lasio's to_json() to go straight into any wellio "from_json()" and vice versa, so I am happy to follow wellio's lead here.

dcslagel:
Just some background info..  A json-schema could both act as the specification to build to_json() and from_json() functions/classes on.  It would also enable testing and validation of the json files.

[...]

PR #111, opened 2015-12-02
Provide JSONEncoder for dumping LASFile to json

kinverarity1:
Provides a fix for #107


kinverarity1:
See #107 for more discussion

PR #112, opened 2015-12-03
Modify how the df is created to avoid for loop

dagrha:


kinverarity1:
3x speed improvement, I like it! Thanks for the help.

I'll review it now.

kinverarity1:
Thank you, this was excellent - it's been merged via #113. I also started a new example notebook just for showing how to use the pandas DataFrame attribute.

dagrha:
The new example notebook is great.

PR #122, opened 2016-02-01
Dev conda environment

oliveirarodolfo:
First attempt to integrate with binder


oliveirarodolfo:
Sorry for this pull, should be on my local master, I still haven't test it enough to bother you with this.

Issue #124, opened 2016-02-02
Refactor LASFile into smaller classes


kinverarity1:
Should have separate classes (or functions) for:

- Reader
- Writer

Also should have a kind of wrapper/mixin to provide special properties for v1.2 and v2 LAS files, like "well", "params", "other".

And rename LASFile to File - LAS is a duplicate given it's in the las module.

Possibly even separate modules: las.py and parser.py.

PR #129, opened 2016-02-13
Project refactor

kinverarity1:
Move things around a bit:

Still very ugly with imports &c. - need to work more on this.

Also create some actual wrapping scripts in a scripts folder.

Ref #124


kinverarity1:
ref #124

PR #130, opened 2016-03-14
Write some good repr methods

kinverarity1:
Better `__str__` and `_repr_html` methods for:
- HeaderItem
- CurveItem
- SectionItems
- LASFile

Need to think about whether I need to improve `__repr__` for LASFile.


kinverarity1:
This is more or less done.

PR #137, opened 2016-07-14
Fix language statistics

roliveira:
GitHub linguistics identified all `*.las` files as LASSO language and added them in the statistics of the repo. This small modification in the `.gitattribute` corrects this, ignoring the `*.las` files when calculating the statistics.

*Jupyter Notebooks* are still being recognised as the majority of the files, but I imagine they would vanish as issues get solved.


kinverarity1:
Thanks!

PR #149, opened 2017-01-19
Read 3.0 with pandas

ahjulstad:
Just sharing my working code that reads some LAS 3.0 files with only a single ~ASCII section.  It works for me, but breaks several write tests. I don't need the write functionality, so. What I needed was the ability to parse datetime fields correctly, and it appears to work for me.

It uses pandas instead of numpy for the data structure and parsing of data. It also requires dateparser (for parsing dates in parameter fields). 

The code is not pretty, but perhaps it might be useful to others with equivalent use cases.

I have based it on latest master, but filed the pull request to the las3.0 branch.


kinverarity1:
Hi - thank you so much for working on this. I'll have a look as soon as I get a chance.

PR #150, opened 2017-02-05
Remove named list from dependencies

nasedil:
Since namedlist is not used, it can be removed from code and as a dependency.  That may help to solve issues with insntalling lasio in latest anaconda3.  Should solve #151.


kinverarity1:
Thanks!

PR #156, opened 2017-06-18
delete_curve method, unit test

dagrha:
Added a method for deleting a curve from a LASFile object. First gets the index in the `curves.keys()` list of the curve argument, then deletes the corresponding CurveItem from the `curves` attribute, then deletes the corresponding "column" in the ndarray `data` attribute. Also added a test-- hopefully it is up to snuff.

Apologies about the STRT, STOP, STEP, and Returns lines. My text editor must have automatically removed trailing spaces.


kinverarity1:
Thanks, this looks great - and thanks so much for adding the unit test too.

PR #157, opened 2017-07-07
Add cases for WRAP and continue scan lines if not match with pattern

desperado1288:
Continue scanning lines if not match with pattern to avoid stopping running when some lines don't match the pattern that doesn't effect the whole file.

### DIFF
--- a/lasio/las.py
+++ b/lasio/las.py
@@ -123,7 +123,13 @@ class LASFile(object):
                 read_parser.version = 1.2
             else:
                 read_parser.version = 2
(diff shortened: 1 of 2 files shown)


kinverarity1:
Thanks for this! I've tried to address the issue in PR #173 - see the comments there.

kinverarity1:
This should have been fixed by PR #173, which is now in master. Thanks for the suggestion! Please re-open the issue if you find it is still crashing on read for you.

Issue #101, opened 2015-08-24
Memory usage for reading of 268 MiB


VelizarVESSELINOV:
My feeling is that lasio is using too much memory. Please find a test program that I wrote to monitor the memory usage: https://github.com/VelizarVESSELINOV/DockerTest

I created a dummy LAS file with 0.5M rows and 51 mnemonics the result is 267.98 MiB file which is reasonable size (without being very large) for LAS file.

My expected maximum memory usage is numpy.ndarray which is most probably 1/8 of the LAS ASCII file.

I didn't expect the raw ASCII file to be put in memory. The result of `pympler` after `lasio.read`:

```
Reading file
                                types |   # objects |   total size
===================================== | =========== | ============
                          <class 'str |        4356 |    268.28 MB
                         <class 'list |        4214 |    395.56 KB
            <class 'lasio.las.LASFile |           1 |     35.26 KB
                          <class 'int |         778 |     21.27 KB
                    <class 'weakproxy |         176 |     13.75 KB
            <class 'collections._Link |         176 |     12.38 KB
                         <class 'dict |           9 |      8.25 KB
  <class 'lasio.las.OrderedDictionary |           3 |      6.43 KB
                <class 'numpy.ndarray |          51 |      3.98 KB
              <class 'lasio.las.Curve |          51 |      3.98 KB
         <class 'lasio.las.HeaderItem |          19 |      1.34 KB
                         <class 'type |           0 |    576     B
                        <class 'float |           5 |    120     B
                        <class 'tuple |           1 |     64     B
      function (_count_diff_hashable) |           0 |      0     B
```

The Mac Activity monitor is giving me memory usage of 3 GiB with 5.3 GiB virtual memory.

<img width="559" alt="screen shot 2015-08-23 at 7 59 46 pm" src="https://cloud.githubusercontent.com/assets/8209940/9432582/12e6830a-49da-11e5-8751-bca5ec8739d4.png">

The Python `memory_profiler` is giving line-by-line memory consumption of 500 MiB per read.
I have the feeling that they are not capturing the numpy allocations that are done in C or Fortan.

```
Line #    Mem usage    Increment   Line Contents
================================================
    31     27.8 MiB      0.0 MiB   @profile
    32                             def file_test(rows=500000, cols=50):
    33                                 "File test"
    34     27.8 MiB      0.0 MiB       print("Creating file with {} rows and {} columns".format(rows, cols))
[...]

Issue #50, opened 2015-08-05
With big LAS files potential memory error


VelizarVESSELINOV:
With some of the big LAS files `numpy.loadtxt(StringIO(s))` is crashing due to not enough memory.

In the past I was using the solution described here http://stackoverflow.com/questions/8956832/python-out-of-memory-on-large-csv-file-numpy

I can generate a big las file, but depends on the memory available in my docker container can crash the parser.

This is an example even before during replace is crashing.

Line by line of chunk by chunk should be used to optimize the memory footprint.

Wrap exampled #1: the STRT is wrongly setup (above Everest altitude) in this example, but is not strange to have files with a lot of rows :)

```
~Well Information Block
STRT.FT           -178957.0000: START DEPTH
STOP.FT              8372.0000: STOP DEPTH
STEP.FT                 0.5000: STEP
~Curve Information Block
DEPT.FT            0 000 00 00: Depth
R01.IN                        : 
R02.IN                        : 
R03.IN                        : 
R04.IN                        : 
R05.IN                        : 
R06.IN                        : 
R07.IN                        : 
R08.IN                        : 
R09.IN                        : 
R10.IN                        : 
R11.IN                        : 
R12.IN                        : 
R13.IN                        : 
R14.IN                        : 
R15.IN                        : 
R16.IN                        : 
R17.IN                        : 
R18.IN                        : 
R19.IN                        : 
R20.IN                        : 
R21.IN                        : 
R22.IN                        : 
R23.IN                        : 
R24.IN                        : 
R25.IN                        : 
R26.IN                        : 
R27.IN                        : 
R28.IN                        : 
R29.IN                        : 
R30.IN                        : 
R31.IN                        : 
R32.IN                        : 
R33.IN                        : 
R34.IN                        : 
R35.IN                        : 
R36.IN                        : 
R37.IN                        : 
R38.IN                        : 
R39.IN                        : 
R40.IN                        : 
R41.IN                        : 
R42.IN                        : 
R43.IN                        : 
R44.IN                        : 
R45.IN                        : 
R46.IN                        : 
R47.IN                        : 
R48.IN                        : 
R49.IN                        : 
R50.IN                        : 
R51.IN                        : 
R52.IN                        : 
R53.IN                        : 
R54.IN                        : 
R55.IN                        : 
R56.IN                        : 
R57.IN                        : 
R58.IN                        : 
R59.IN                        : 
R60.IN                        : 
idmx.IN                       : 
idmn.IN                       : 
[...]

PR #159, opened 2017-07-23
Improve memory and speed footprint

kinverarity1:
This PR will once finished read LAS files in by iteration, which should increase speed and significantly reduce the memory overhead (see #50 and #101, and [benchmarking code](https://gist.github.com/kinverarity1/e24762c17da6a7643c829e6ba2168ecd).

### DIFF
--- a/lasio/reader.py
+++ b/lasio/reader.py
@@ -83,7 +83,27 @@ def open_file(file_ref, encoding=None, encoding_errors='replace'):
     return file_ref
 
 
-def read(file_ref, null_subs, **kwargs):
+def read(file_ref, null_subs=True, **kwargs):
+    '''Read file contents into memory.
+
+    Arguments:
+        file_ref: either a filename, an open file object, or a URI.
+
+    Keyword Arguments:
+        null_subs (bool???): ????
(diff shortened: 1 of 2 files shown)


lasio/reader.py
```diff
@@ -68,28 +66,61 @@ def open_file(file_ref, encoding=None, encoding_errors='replace',
...
+def read(file_ref, null_subs, **kwargs):
+    sections = OrderedDict()
+    sect_lines = []
+    sect_title_line = None
+    file_ref = open_file(file_ref, **kwargs)
+    for line in file_ref:
+        line = line.strip()
+        if line.startswith('~A'):
+            # We have finished looking at the metadata and need
+            # to start reading numerical data.
```
dagrha: If a user were interested in _only_ the header data in a LAS file, this could be a good place to add a `header_only` option. If `True` it would not read in the log data. In other words, could be something like ```if not header_only: 
    do stuff with ~A section```
Not having to read in all the log data would significantly speed things up if the user is only interested in header data. Just a thought. Cheers.

kinverarity1:
Superseded by #170

PR #170, opened 2017-10-27
Iterative read v2

kinverarity1:
This long-overdue PR implements an iterative read. Instead of reading the whole LAS file into memory, ``reader.py`` now looks at the file only line by line - the number of passes has gone from who knows how many to a maximum of 1 + n, where n is the number of data sections.

There are two new features:

1. ``lasio.read`` has a new keyword argument ``ignore_data`` (thanks @dagrha for [suggesting this](discussion)). If this is True, then lasio will not read the file beyond the start of the ~ASCII data section. Attempting to use the ``LASFile.data`` attribute, or any methods that use it (many!) will raise an ``AttributeError``.
2. ``lasio.read`` also has a new keyword argument ``ignore_header_errors``. Anything that would have raised a exception (``lasio.exceptions.LASHeaderError``) will instead result in the same message being sent to ``logging.debug`` and ignored. The remaining parts of the header should successfully be parsed and loaded.
    (Note this involved changing one unit test, so you may need to change production code to accommodate this with ``ignore_header_errors=True`` if have any poorly-formatted header lines in the Parameter section. It is set to False by default.)

And performance improvements ([benchmarking code and results](https://github.com/kinverarity1/lasio-speed)):

- Memory usage. For an example 5 MB LAS file, the memory usage has gone from [30 MB](https://github.com/kinverarity1/lasio-speed/blob/master/result-3b47a23.txt#L16) to [3.3 MB](https://github.com/kinverarity1/lasio-speed/blob/master/result-4d6678c.txt#L16). 
- Speed. For the same example, load time went from [7.6 s](https://github.com/kinverarity1/lasio-speed/blob/master/result-3b47a23.txt#L19) to [1.9 s](https://github.com/kinverarity1/lasio-speed/blob/master/result-4d6678c.txt#L19) on my machine. Still not great, but much better than it was.

### DIFF
--- a/lasio/__init__.py
+++ b/lasio/__init__.py
@@ -27,6 +27,9 @@ def read(file_ref, **kwargs):
             a LAS file contents.
 
     Keyword Arguments:
+        ignore_data (bool): if True, do not read in any of the actual data, just
+            the header metadata. False by default.
+        ignore_header_errors (bool): ignore lASHeaderErrors: False by default
         encoding (str): character encoding to open file_ref with
         encoding_errors (str): 'strict', 'replace' (default), 'ignore' - how to
             handle errors with encodings (see standard library codecs module or
--- a/lasio/defaults.py
+++ b/lasio/defaults.py
@@ -1,3 +1,4 @@
+import re
 
 import numpy as np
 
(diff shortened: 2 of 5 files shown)


dagrha:
Damn, Kent! Closing all sort of tickets! Looks like a great PR. Kudos and thanks are in order :+1:

kinverarity1:
Thanks - long overdue (and thanks @VelizarVESSELINOV for originally suggesting this).

Issue #166, opened 2017-10-24
Degree symbol is making reader break


EvanBianco:
In the `~Parameter Information` section of the attached here, the `reader.py` is throwing an error presumably when it encounters the degree symbol in the latitude and longitude fields. 

```
LATI .        0° 0' 0  N                 :Latitude
LONG .        0° 0' 0  E                 :Longitude
```

The error message I'm getting has something to do with codecs. Is this something we can fix with the reader or it a problem that has to do with my Python installation?

[1045387215_top100ft.txt]
(https://github.com/kinverarity1/lasio/files/1410966/1045387215_top100ft.txt)

kwinkunks:
Here's the traceback...

```
>>> l = lasio.read('/Users/matt/Downloads/1045387215_top100ft.txt')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/matt/anaconda/envs/welly/lib/python3.5/site-packages/lasio/las.py", line 1269, in read
    return LASFile(file_ref, **kwargs)
  File "/Users/matt/anaconda/envs/welly/lib/python3.5/site-packages/lasio/las.py", line 378, in __init__
    self.read(file_ref, **kwargs)
  File "/Users/matt/anaconda/envs/welly/lib/python3.5/site-packages/lasio/las.py", line 403, in read
    f = open_file(file_ref, **kwargs)
  File "/Users/matt/anaconda/envs/welly/lib/python3.5/site-packages/lasio/las.py", line 1059, in open_file
    autodetect_encoding_chars)
  File "/Users/matt/anaconda/envs/welly/lib/python3.5/site-packages/lasio/las.py", line 1105, in get_unicode_from_filename
    data = f.read()
  File "/Users/matt/anaconda/envs/welly/lib/python3.5/codecs.py", line 321, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb0 in position 3243: invalid start byte
>>
```

kwinkunks:
OK, this works... 

    l = lasio.read('/Users/matt/Downloads/1045387215_top100ft.txt', encoding='utf-8')

I guess the default is... ASCII? (Imagine that!)

kwinkunks:
I guess the problem is there are no BOM chars. I wonder if a good approach would be to put a try-except around it and attempt to read with a couple of common codecs.

[...]

PR #174, opened 2017-10-28
Fixing #166

dagrha:
On Ubuntu 17.10 even after #170 I was still getting UnicodeDecodeError on the file referenced on #166, I think because I was specifying neither that `autodetect_encoding=True` nor `encoding='blah-blah'`, like so:

    l = lasio.read('/home/dagrha/Downloads/1045387215_top100ft.txt')

The program was not seeing a BOM but it was also not being told to do autodetect_encoding, so the UnicodeDecodeError was still being raised.

As discussed in https://github.com/kinverarity1/lasio/issues/166#issuecomment-339321558 I've changed the logic a little bit. First, I changed `autodetect_encoding` to be True by default. If chardet is not installed though, then a few common encodings are tried. If that fails, then finally UnicodeDecodeError is raised.

I'm not sure that what I've got here is the best solution, but it is now working for me. I still need to write some tests for this.

I apologize about all those spurious deletions and insertions the my editor put in there.


kinverarity1:
Thanks for this! I had a look and realised this is still tricky. I think the behaviour of ``autodetect_encoding`` should depend on what the user does with the ``encoding`` argument - otherwise a user might be under the assumption that the encoding they specified is correct, even though lasio found it raised a UnicodeDecodeError, but managed to open it with a different one. If the user has logging off (which 99% will), then they'll never know.

So the simple solution I'm writing up now is to get rid of autodetect_encoding in a backwards-compatible way, and add 'auto', 'chardet', and 'cchardet' options to the encoding argument. Then, whenever a user doesn't specify an actual encoding, lasio tries the common list as you have put in here.

and still feel a bit uncomfortable with the logic that if autodetect_encoding=False, it would still be testing a variety of encodings. i.e. if the user specifies ``read(autodetect_encoding=False, encoding='latin-1')``, has logging off, and the actual encoding is windows-1252, then lasio would have successfully parsed the file as windows-1252 and the user might be under the reasonable assumption that the actual encoding is indeed say latin-1 - even though ``codecs.open(..., encoding='latin-1')`` was raising a UnicodeDecodeError

Anyway I've kept your changes and made some more so that even autodetec

kinverarity1:
Sorry - just realised your PR doesn't lead to the situation I described. Never mind. Thanks!

kinverarity1:
I needed a branch here to push to, so am working on your code in #175  sorry

dagrha:
Ooh, I like your idea about just having an `encoding` argument.

Just as a discussion point, would it make any sense to make `encoding='auto'` the default behavior, i.e. when encoding is not explicitly specified in `read()`? The only reason I ask is that relying on chardet could be safer bet for guessing the correct encoding. Then `adhoc_test_encoding` (i.e. list of common encodings) would only be used if chardet is not installed. Or does relying on chardet by default add too much overhead? 

Either way, fine by me, just wanted to throw that out there!

kinverarity1:
Yep I think that sounds good, i think people prefer error-free posting by default!! I'd just make chardet and maybe cchardet dependencies for lasio. Seems worthwhile.

PR #199, opened 2017-11-29
add logging level notebook and example las

dagrha:
I keep having to re-lookup how to change change the logging level when using lasio, so I thought I'd throw together a [jupyter notebook](http://nbviewer.jupyter.org/github/dagrha/lasio/blob/master/notebooks/set%20logging%20level%20for%20lasio.ipynb) for reference. If it's something you think is useful, feel free to merge it, otherwise no worries if not! I also added an example LAS file to the tests/examples directory that the notebook uses.

### DIFF
--- a/notebooks/set logging level for lasio.ipynb
+++ b/notebooks/set logging level for lasio.ipynb
@@ -1,0 +1,242 @@
+{
+ "cells": [
+  {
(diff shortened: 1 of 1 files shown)


kinverarity1:
Thank you! :+1:

PR #238, opened 2018-05-21
Allowing characters in the data section

krzysztof-slowinski:
A small example to approach allowing the characters to be present in the data section.

Leaving the values that could not be parsed as float64 as strings.

### DIFF
--- a/lasio/reader.py
+++ b/lasio/reader.py
@@ -349,9 +349,12 @@ def read_data_section_iterative(file_obj, regexp_subs, value_null_subs):
             for pattern, sub_str in regexp_subs:
                 line = re.sub(pattern, sub_str, line)
             for item in line.split():
-                yield item
+                try:
+                    yield np.float64(item)
+                except ValueError:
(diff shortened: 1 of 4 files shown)


kinverarity1:
Thanks, this is a great start. We may need to look at the test errors on py 3.3 and 3.4 - https://travis-ci.org/kinverarity1/lasio/builds/381735089?utm_source=github_status&utm_medium=notification

Also need to ensure memory usage remains low - [L357](https://github.com/kinverarity1/lasio/pull/238/files#diff-5032f0bd21315b89d455b27f86700356R357) worries me with replacement of ``numpy.fromiter`` with a list comprehension which I think may be a large memory usage for large i.e. 200 MB+ LAS files (``np.array([i for i in items(file_obj)])``).

kinverarity1:
Maybe can check memory usage with benchmarking code linked to from #170

kinverarity1:
had to shift to #245

PR #239, opened 2018-05-25
Update writer.py

Fry484:
When loading a .las into Petrel 2016, the .las importing utility will not recognize the curves with, "~Curves," but does recognize the curves with, "~Curve" or "~Curve Information"


kinverarity1:
Thanks for this @Fry484  - the tests just needed updating; see #240

PR #249, opened 2018-07-13
after allowing characters in the data section, df() function was retu…

krzysztof-slowinski:
…rning DataFrame with columns of dtype object for all columns if there are characters in the data section - this is a fix by attempting to convert the columns to float64 if they only contain numerical data

### DIFF
--- a/lasio/las.py
+++ b/lasio/las.py
@@ -557,7 +557,14 @@ class LASFile(object):
     def df(self):
         '''Return data as a :class:`pandas.DataFrame` structure.'''
         import pandas as pd
+        from pandas.api.types import is_object_dtype
         df = pd.DataFrame(self.data, columns=[c.mnemonic for c in self.curves])
+        for column in df.columns:
+            if is_object_dtype(df[column].dtype):
+                try:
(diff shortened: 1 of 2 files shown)


kinverarity1:
Thanks! I am away on some fieldwork at the moment but will try to merge this asap when I get a chance.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

krzysztof-slowinski:
It is worth to mention that I tried to do the fix in las.py#set_data where the curve.data is assigned (currently at line 623), but fixing it at that location did not result in the change of type in the DataFrame returned through df() due to the way this DataFrame is constructed. If possible, it would be worth to fix it at the origin, which las.py#set_data seems to be, but it would require some additional changes.

kinverarity1:
OK I see what you mean. It'll be a bit messy but I'll add some code in LASFile.set_data_from_df to try and convert each DataFrame column to floats if possible. In addition to your commit.

Thanks!

kinverarity1:
I'll merge this now and work on what I mentioned when possible. Cheers.

PR #258, opened 2018-10-11
Make Reader Keyboard-Interruptable and more robust for faulty LAS files

bgschaid:
I use the library to batch-process some LAS-files. When trying to interrupt the script with Ctrl-C it is not stopped because the `KeyboardInterrupt` is caught by a general `except:`and reraised as a `LASDataException` which definitely is not the right behaviour.

The first commit catches the Keyboard-Interrupt and reraises it

The second commit is only trailing whitespaces that my editor removes automatically. It can be ignored

General remark: the catch-all `except:` is not good practice and should always be replaced with the concrete exceptions that are to be expected

### DIFF
--- a/lasio/__init__.py
+++ b/lasio/__init__.py
@@ -1,5 +1,8 @@
 import os
 
+import logging
+logger = logging.getLogger(__name__)
+
 from .las import LASFile, JSONEncoder
 from .las_items import CurveItem, HeaderItem, SectionItems
 from .reader import open_file
@@ -10,7 +13,6 @@ except ImportError:
     pass
 else:
     from .excel import ExcelConverter
-
 
 __version__ = '0.22'
 
--- a/lasio/excel.py
+++ b/lasio/excel.py
@@ -39,7 +39,7 @@ class ExcelConverter(object):
         Two sheets are created:
 
         * Header: contains all the header sections and metadata
-        * Curves: contains the data 
+        * Curves: contains the data
 
         '''
         wb = openpyxl.Workbook()
@@ -109,7 +109,7 @@ def main():
 
 
 def get_parser():
-    parser = argparse.ArgumentParser('Convert LAS file to XLSX', 
+    parser = argparse.ArgumentParser('Convert LAS file to XLSX',
         formatter_class=argparse.ArgumentDefaultsHelpFormatter)
     parser.add_argument('LAS_filename')
     parser.add_argument('XLSX_filename')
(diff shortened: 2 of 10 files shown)


bgschaid:
Wanted to generate a new pull request for the next 6 request but it seems that github adds them to this. Will try to fix the broken tests

kinverarity1:
Thanks for this! I should be able to have a look on the weekend.

Cheers

Sent from Nine<http://www.9folders.com/>
________________________________

bgschaid:
OK. The last commit fixes the tests. Basically: an exception does not occur anymore

Most commits are independent of each other except for https://github.com/kinverarity1/lasio/commit/c1d136bf1d857af20d4122f32243d528e34e3bc7 on which all the other commits depend

All problems that are found during reading the LAS file (including those that "only" issued a warning until now) add an item to a `.problems` list that holds the warning texts and some additional information (that is not very systematically)

The motivation was that
- the parser is more robust if there are problems and gives back all information that **can** be parsed
- information about problems is available (including the number of problems)

kinverarity1:
Thanks so much for this work! I am happy to implement these but I think any changes to the behaviour of the parser like ca1bcf8, 1606420 and 02ac8e4 should each come with a specific regression test. I am happy to do these when I have time, but would welcome any regression tests you could include. Really all it needs to be is a LAS example file that used to "fail", but with your code change, passes. Preferably each of those commits should be a separate PR.

Please leave this PR open though, I will incorporate the idea of a "problem" list and merge it.

bgschaid:
I've added the requested tests (used one of your test files and edited it to resemble problems I've seen in LAS-files)

Also added asserts for the number of problems found by the reader to all reader tests

kinverarity1:
Thanks for the contribution! Sorry about taking a year to get to it.

I've merged your fix about Ctrl+C not being caught - thanks for that.

The others I think deserve perhaps their own consideration, so I will close this PR for now. Please feel free to raise a new issue or PR to discuss - some initial thoughts here:

- 02ac8e4: Lines that do not match the pattern of a proper header line do not cause a failure but are added as UNMATCHED and reported as problems - I would prefer to parse them somehow and include them in the header section.. not sure about the idea of counting an "unmatchedNbr" variable
- 1606420: Detect if sections are duplicated in the file and add these to the problems - I'd prefer to just keep the section names as they are, similar to how we handle duplicate curve names
- ca1bcf8: Catch problems were the ~A section is not read correctly because of some misformatting but make the data that COULD be read available - I like the idea but unless I'm missing something, that's not what this commit does? I've improved the error message in #292. To read only the header, I think a more reliable pattern is to raise the error so that client code can pick that up and if they choose, they can re-read the file with the `lasio.read(ignore_data=True)` kwarg.
- c1d136b: Collect all the problems in the file that cause warnings in a list for later analysis instead of just printing them - I'd prefer to do this using the Python warnings module, and leave the collection of messages up to client code e.g. https://docs.python.org/2/library/logging.html#logging.captureWarnings

Issue #36, opened 2015-08-01
Period following unit in header items 


kinverarity1:
CDD1.IN_FT.                225: CASING DRILLER

Doesn't seem very sensible and fairly easy to fix.

PR #283, opened 2019-07-25
Added check_conforming function

MandarJKulkarni:
Added first version of check_conforming function to check missing mandatory sections ~version and ~well.


kinverarity1:
Thanks for the pr - see discussion in #36

kinverarity1:
Before we look at merging this, I think we need to make it do complete I.e. check all the requirements of the LAS spec. This would be a really neat feature but unfortunately it's an all-or-nothing sort of thing. We don't want to be returning `True` for files that have problems.

We could of course rename your method like you suggest but then later the API will become cluttered, which I definitely don't want.

lasio/las.py
```diff
@@ -64,6 +64,8 @@ def __init__(self, file_ref=None, **read_kwargs):
         super(LASFile, self).__init__()
         self._text = ""
         self.index_unit = None
+        self.version_section_not_found = False
+        self.well_section_not_found = False
```
kinverarity1: Avoid double negatives e.g. `version_section_exists = True` is better

lasio/las.py
```diff
@@ -139,6 +141,11 @@ def add_section(pattern, name, **sect_kws):
                 logger.warning(
                     "Header section %s regexp=%s was not found." % (name, pattern)
                 )
+                if name == "Version":
+                    self.version_section_not_found = True
+                elif name == "Well":
+                    self.well_section_not_found = True
+
```
kinverarity1: Don't do the check just on the LAS file being loaded - the user might add sections later. The check should be done every time ``check_conforming()`` is called.

MandarJKulkarni: The catch is, if a section doesn't exist the default value added in __init__ i.e. default_items["Version"], default_items["Well"] etc. persists. So should we delete the default value/s for sections in such case? Because otherwise the las file will always have version, well sections.

kinverarity1: Good point. I think the best solution is to only create the default sections if we are **not** reading a file in. e.g. something like cc0430. That commits breaks some tests, but I'm not worried. At the moment we are producing side-effects by reading a file in which is pretty odd.

kinverarity1: Ultimately I want to get to a place where `lasio.read("blah.las").write()` produces something identical, and `lasio.read("blah.las").to_conforming().write()` makes the necessary changes like added sections, cleaned up formatting, etc.

MandarJKulkarni: > Good point. I think the best solution is to only create the default sections if we are **not** reading a file in. e.g. something like [cc0430b](https://github.com/kinverarity1/lasio/commit/cc0430bb1e59bee1459de2ac533508a641f48aed). That commits breaks some tests, but I'm not worried. At the moment we are producing side-effects by reading a file in which is pretty odd.

OK, I'll use this approach then. I'll fix those tests or remove if required.

MandarJKulkarni:
I think I'll close this PR and create a new one.

lasio/las.py
kinverarity1: No worries. Thanks for working on this! 

I've looked at the failing tests. Don't worry about `test_write_wrapped` or `test_write_unwrapped` - I'll fix those later, they are poorly written. 

I'd like to have `test_barebones` and  `test_barebones_missing_all_sections` kept as they are, I can work on a fix for that too so don't worry about them continuing to fail.

MandarJKulkarni:
These changes have the checks for required sections only, are there any other checks we should put? Please let me know and other comments as well if any.

tests/test_read.py
```diff
@@ -239,7 +240,7 @@ def test_barebones():
 
 def test_barebones_missing_all_sections():
     las = lasio.read(egfn("barebones2.las"))
-    assert las.curves[-1].mnemonic == "UNKNOWN:8"
+    assert las.sections == {}
```
kinverarity1: This needs to go back to the old test, I want to find a way to retain this behaviour -- I'll work it out.

tests/test_read.py
```diff
@@ -252,6 +253,21 @@ def test_check_conforming_no_well():
...
+
+
+def test_check_conforming_no_ascii():
+    las = lasio.read(egfn("missing_ascii_section.las"))
+    assert not las.check_conforming()
+
+
+def test_check_conforming_no_ascii_2():
+    las = lasio.read(egfn("missing_curves_section.las"))
+    assert not spec.AsciiSectionExists.check(las)
```
kinverarity1: ?? This test should be failing: `missing_curves_section.las` contains an ~A section.

lasio/las.py
```diff
@@ -264,16 +260,17 @@ def add_section(pattern, name, **sect_kws):
...
+                    n_curves = len(self.curves)
+                    n_arr_cols = len(self.curves)  # provisional pending below check
+                    logger.debug("n_curves=%d ncols=%d" % (n_curves, s["ncols"]))
+                    if wrap == "NO":
+                        if s["ncols"] > n_curves:
+                            n_arr_cols = s["ncols"]
+                    data = np.reshape(arr, (-1, n_arr_cols))
+
+                    self.set_data(data, truncate=False)
+                    drop.append(s["title"])
```
kinverarity1: See comment in main PR thread.

MandarJKulkarni: Which comment?

MandarJKulkarni:
Hi, Should I create a separate PR for the change to create default sections only if we are NOT reading a file? or would that break backwards-compatibility?

lasio/las.py
kinverarity1: Sorry! This is my first time using Github's PR-reviewing function, so I am a bit lost ;-) I'm referring to https://github.com/kinverarity1/lasio/pull/283#pullrequestreview-270520602. I think I will try and tackle this part, as it's cleaning up my mess from the past :-p

kinverarity1:
Hey, nah don't worry about splitting it into separate PRs. Let's keep on with this one.

lasio/las.py
MandarJKulkarni: Ah I see, sorry I also couldn't co-relate. 
if you want to read the file with only **data** sections
One approach I could think of is to change the set_data function to check if there are no Curves and then set the array value as it is for las.data (new attribute) if there are no Curves. This could be a quicker fix, however the data returned wouldn't be always very readable.

tests/test_read.py
MandarJKulkarni: I added this to reflect current behavior of `spec.AsciiSectionExists.check`, where ascii sections existence is checked through existence of data in Curves section only.
So if the Curves section is missing, there is no way to access the data or ascii section.
This ties back to your earlier comment where you want to support reading files like `barebones2.las` where only data sections exist.

MandarJKulkarni: I can delete the test and we can add something similar later

MandarJKulkarni:
The current code assumes that ~A is always the last section in the file and hence it is trickier to detect duplicate ~A in the file (and also other duplicate sections if they appear after ~A section).

I see 2 ways to fix this:
1. Don't treat ~A section as the last section in the file, but do mark it as data section, and read arrays using numpy separately.
2. Modify read_data_section_iterative to indicate duplicate sections, if any of the lines in the section start with '~'.

I am trying to code approach 1, but let me know if you think approach 2 is better and also if there is any other better approach.

kinverarity1:
Thanks for all the work on this, looking good! I think option 3 is best for now, it's a good idea. The long-term plan is to read all non-data sections, parse them, then return to the data sections to read (see discussion in #1).

MandarJKulkarni:
Hi, do you think there should be more checks that we need to add? Do you see any obvious ones that I've missed?

kinverarity1:
I'm short on time at the moment but the next step for me would be to go through the [v1.2 and v2.0 standards](https://github.com/kinverarity1/lasio/tree/master/standards) and highlight the passages which sound like a requirement.

MandarJKulkarni:
Hi,
I am currently only looking at LAS 2.0 standards and I think this is the
status.


Below are the yet-to-do as per 2.0

The depth value divided by the step value must be a whole number

The index curve (i.e. first curve) must be depth, time or index. The only
valid mnemonics for the
index channel are DEPT, DEPTH, TIME, or INDEX.

Time and date can be included in LAS 2.0 files provided that they are
expressed as a number.

"~V" must be the first section.


Embedded blank lines anywhere in the section are forbidden

And for below, code changes are present in the PR.

"~V" is a required section;

 "~W" (also known as "WELL INFORMATION SECTION") is a required section;

*"~C" *(also known as ~CURVE INFORMATION SECTION") is a required section;

*"~A" *(also known as ~ASCII LOG DATA") is a required section;

Only one *"~V" *section can occur in an LAS 2.0 file.

~V section must contain the following lines.

Only one *"~W" *section can occur in an LAS 2.0 file.

~W section must contain the following lines.

Only one *"~C" *section can occur in an LAS 2.0 file.

Only one *"~P" *section can occur in an LAS 2.0 file.

Only one *"~O" *section can occur in an LAS 2.0 file.

The data section will always be the last section in a file.

Please let me know what you think.

Thanks,
Mandar J Kulkarni.

kinverarity1:
Thanks for all your work, and this list - I am really short on time at the moment, but I will certainly have a close look and hopefully soon merge and release this.

MandarJKulkarni:
Hey, Sure! Thanks, do let me know your comments.

MandarJKulkarni:
I have added checks for below in the last 3 commits.

- The depth value divided by the step value must be a whole number

- The index curve (i.e. first curve) must be depth, time or index. The only
 valid mnemonics for the index channel are DEPT, DEPTH, TIME, or INDEX.

- "~V" must be the first section.

MandarJKulkarni:
Hi Kent, Please have a look and let me know your comments.

kinverarity1:
Thanks for all this work! I've had a closer look and think we might need to consolidate some of the changes - I need some time to review this fully and get back to you.

MandarJKulkarni:
> Thanks for all this work! I've had a closer look and think we might need to consolidate some of the changes - I need some time to review this fully and get back to you.

sure Kent, do let me know.

MandarJKulkarni:
Hi Kent, now I have added checks for all the rules for LAS2.0 from the list I had prepared.
I see you have listed the rules for LAS1.2.

Maybe the function check_conformity should check the version of LAS file first and then apply rules accordingly. What do you suggest?

MandarJKulkarni:
Hi Kent,

Sorry to bother you again,did you get a chance to look at the changes?

Thanks,
Mandar

kinverarity1:
Hey @MandarJKulkarni I've had very little time but will keep working through them through December. Yep, I'm still working out how to handle both 1.2 and 2.0 together as some of the rules are unchanged between the standards.

dcslagel:
@kinverarity1,

I read through the code in this change and am interested to help it progress.
Is there some sub-section of  work I could do to help move this PR forward?

Thanks,

DC

kinverarity1:
Sure thing! I'm working on the framework for how I'd like to implement this in the ``check-conforming`` branch, so it would be best to wait for a day or two until I can finish that up and then I suspect there will be a reasonable amount of effort involved in implementing some of the checks, as we might have to adapt and improve the reader code.

Thanks @MandarJKulkarni for getting this started!

dcslagel:
Okay, great,  Yep when you are ready let me know where I might help in the effort.

Thanks,

DC

MandarJKulkarni:
Hi Kent,
I was looking at the branch check-conforming. Can you explain more about "requirements", the argument to method 'conformity'?

Also I was looking at checks.py, do you want to have version specific checks, like WellSectionExists_V12 and WellSectionExists_V20? But the content of both would be same?

Instead, within conformity, can we determine the version first and then call specific set of checks?
So we wouldn't have duplicate code for the checks which are applicable for both versions:
if "version" == "1.2":
     version12_checks()
elif "version" == "2.0":
     version20_checks()
else:
     fail

kinverarity1:
Hi all! I think there's a lot of valuable work in this branch, but I've been thinking a lot about it and am feeling I would prefer to shift this work to a different package.

I apologise for not making it clear earlier, but I would like lasio to remain focused on the task of parsing files and providing easy access to all the metadata and data in a LAS file. I think that is all that I have time to commit to (in terms of maintenance and code review, etc.) Unfortunately I don't think the task of checking whether a LAS file meets certain standards is within that scope.

@MandarJKulkarni I'm aware I have held up the valuable work that you done in this branch, but I hope you understand that I don't want lasio to contain a feature that isn't complete or doesn't work as users might expect, and more importantly, one that I won't be able to support in the future.

I was planning to shift this branch into a separate repository and Python package e.g. "lascheck"? and provide a structure to it, but I unfortunately have not had time even for that! So I am quite happy for either @MandarJKulkarni or @dcslagel to take that on, if you want? I will also be happy to respond to lasio PRs where new parsing features are required to support functionality in a "lascheck" package.

Thanks for all your efforts here!

MandarJKulkarni:
Hi Kent, I understand your concerns and why you don't want to merge.
The idea of lascheck sounds interesting and I will work towards it. Your guidance (as and when time permits you) in that work will definitely be helpful.
@dcslagel , I will create a repository and let you know, you are welcome to contribute.

PR #288, opened 2019-09-30
Issue278 apu/uwi leading zeros - Add tests and proposed fix

dcslagel:
Here is a proposed fix for :     
Issue 278: Header Attribute API/UWI with Leading Zeros are misread.

It adds two tests.  All 164 tests pass on python 3.7.4

The solution is for Las file versions 2.0 like the example in the bug report.


dcslagel:
Okay, Added changes and tests to also handle the case where the key_order is "descr:value.

dagrha:
This is great, @dcslagel. Thank you for this fix and for adding tests as well!

PR #286, opened 2019-08-13
Fixed error on writing single depth (edge case use)

trqmorgan:
Fixes error rthat occurs calculating STEP if single data point is being written.

PR #289, opened 2019-10-15
Add test case for issue286 - write on single depth

dcslagel:
This pull request contains a test case for the #286 pull request: Fixed error on writing single depth (edge case use) #286.  If accepted, maybe integrate them at the same time.

This pull request introduces a test file with a single depth and 0.000 step.  

Without the #286 pull content, the new test case will fail with 'E           IndexError: index 1 is out of bounds for axis 0 with size 1'

With the #286 pull content the test case will pass.

test command:
pytest tests/test_write.py::test_write_single_step


kinverarity1:
Thank you for contributing this, much appreciated! I have merged via #290 (because I messed up with git!)

Issue #284, opened 2019-07-26
[request] - Add support for borehole images


manuelblancovalentin:
Even though it is possible to use lasio to import borehole images, it requires an extra step of post-processing to merge all channels with the same mnemonic together, which is undesirable.

I.e, a certain borehole image (such as the raw amplitude from a UBI tool) might have N cols (thus ranging from AWBK[0] to AWBK[N]). If I'm not mistaken, currently your code recognizes these entries as separate logs, so it's required to merge them after importation.

It would be awesome If you could just detect that AWBK[0] and AWBK[i] belong to the same log.

Thanks!

manuelblancovalentin:
Btw, I am not sure how difficult that is to implement right now, but I'd be more than happy to contribute and maybe fix this issue, If you wish so.

ThomasMGeo:
It's super hard to do, look into https://github.com/equinor/dlisio

kinverarity1:
This is definitely something we should do - I've seen those kind of files before for broadband sonic.

I'll have a think about the API but implementing it is easy, just index the LASFile.data ndarray with a list instead of a integer.
________________________________

dagrha:
Would making this a method of the LASFile object make any sense (rather than doing it automatically on read)? Something like `las.concatenate_channels()`.

It would be straightforward to do the columns-to-array operation(s) on a DataFrame (`las.df()`), then set again with the `set_data_from_df()` method. 

Maybe the array would get dropped on write (since those aren't part of LAS 1.2/2 spec...or maybe they are?), in other words only the AWBK[1] etc. channels would get written

kinverarity1:
A method makes sense to me, yeah. I prefer "curve" to "channel". Also I wonder if `las.concatenate()` might confuse with anyone trying to join two LAS files together (e.g. `pd.concat()`). Would `las.stack_channels()` or just `las.stack()` make sense? Returning a 2D ndarray.

The LAS file object itself would stay unaffected, since I'm pretty sure in the specs curves can only be 1D.

We could do it on the dataframe but probably also just using indexing on `las.data`, using the indices of each curve e.g. `las.data[:, np.array([3,4,...,103])]` if the first three columns were not part of the image.

dagrha:
Those are all good points. I'd be happy to take a stab at adding this feature.

dagrha:
[...]

PR #293, opened 2019-10-21
stack_curves method added to LASFile object

dagrha:
This PR adds a method to the LASFile object for stacking multi-channel log data as requested in issue #284. 

The method can be fed a curve mnemonic list or a "stub" (a term i've seen used in pandas). If a curve mnemonic list is fed to the method, those curves will be stacked. If a "stub" is supplied, for example `stub='AWBK'`, the method will retrieve all curves with that prefix and stack those. 

The method sorts the curve mnemonics using "natural sorting" (`1, 2, 3, 4, 5, 6, 7, 8, 9, 10` instead of `1, 10, 2, 3, 4, 5, 6, 7, 8, 9`). Sorting can be turned off with `sort_curve=False`.

Also added some tests and example logs.

### DIFF
--- a/lasio/las.py
+++ b/lasio/las.py
@@ -2,6 +2,7 @@ from __future__ import print_function
 
 # Standard library packages
 import codecs
+from collections.abc import Sequence
 import csv
 import json
 import logging
@@ -720,16 +721,20 @@ class LASFile(object):
         Returns:
             2-D numpy array
         """
+        if isinstance(mnemonic, np.ndarray):
+            mnemonic = list(mnemonic)
+
         if (not mnemonic) or (not all([i for i in mnemonic])):
             raise ValueError('`mnemonic` must not contain empty element')
 
         keys = self.curves.keys()
         if isinstance(mnemonic, str):
             channels = [i for i in keys if i.startswith(mnemonic)] or [mnemonic]
-        elif isinstance(mnemonic, list):
-            channels = mnemonic
+        elif isinstance(mnemonic, Sequence):
+            channels = list(mnemonic)
         else:
-            raise TypeError('`mnemonic` argument must be str or list')
+            raise TypeError('`mnemonic` argument must be string or sequence')
+        print(channels)
 
         if not set(keys).issuperset(set(channels)):
             missing = ', '.join(set(channels).difference(set(keys)))
--- a/lasio/las.py
+++ b/lasio/las.py
@@ -2,7 +2,12 @@ from __future__ import print_function
 
 # Standard library packages
 import codecs
-from collections.abc import Sequence
+
+try: # will work in Python 3
+    from collections.abc import Sequence
+except ImportError: # Support Python 2.7
+    from collections import Sequence
+
 import csv
 import json
 import logging
--- a/lasio/las.py
+++ b/lasio/las.py
@@ -704,6 +704,43 @@ class LASFile(object):
                 str(name) for name in df.columns.values
             ]
         self.set_data(df_values, **kwargs)
+    
+    def stack_curves(self, curve_list=None, stub=None, sort_curves=True):
+        """Stack multi-channel curve data to a numpy 2D ndarray. Provide a 
+        list of curves or a stub name (prefix shared by all curves that will
+        be stacked). If `stub` and `curve_list` are both supplied, `stub`
+        is used.
+        
+        Keyword Arguments:
+            stub (str): Supply the first several characters of the channel
+                set to be stacked. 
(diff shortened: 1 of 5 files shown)


lasio/las.py
```diff
@@ -704,6 +704,43 @@ def set_data_from_df(self, df, **kwargs):
...
         self.set_data(df_values, **kwargs)
+    
+    def stack_curves(self, curve_list=None, stub=None, sort_curves=True):
+        """Stack multi-channel curve data to a numpy 2D ndarray. Provide a 
+        list of curves or a stub name (prefix shared by all curves that will
+        be stacked). If `stub` and `curve_list` are both supplied, `stub`
+        is used.
+        
+        Keyword Arguments:
+            stub (str): Supply the first several characters of the channel
```
kinverarity1: I like the stub argument. Looks great.

lasio/las.py
```diff
@@ -704,6 +704,43 @@ def set_data_from_df(self, df, **kwargs):
...
+                set to be stacked. 
+            curve_list (list): Or a list of the curves to be stacked
```
kinverarity1: Clarify whether the list must contain mnemonics, or whether CurveItems can be used too.

lasio/las.py
```diff
@@ -704,6 +704,43 @@ def set_data_from_df(self, df, **kwargs):
...
+            2-D numpy array
+        """
+        curves = self.curves.keys()
+        if stub:
+            channels = [i for i in curves if i.startswith(stub)]
+        elif curve_list:
+            channels = curve_list
+        if sort_curves:
+            channels.sort(key=self._sort_channels)
+        if channels:
```
kinverarity1: Instead of if blocks here and at line 733 I feel we should assume that channels and therefore indices are found. If not the function should raise an appropriate exception, rather than returning None. We don't necessarily need to raise it but hopefully it will raise something sensible if we drop the if blocks.

Also that case - where channels are not found - should also be a test.

lasio/las.py
```diff
@@ -704,6 +704,43 @@ def set_data_from_df(self, df, **kwargs):
...
+            indices = [curves.index(i) for i in channels]
+        if indices:
+            return self.data[:, indices]
+    
+    def _str_to_int(self, mnemonic):
```
kinverarity1: Doesn't feel like this shouldn't be a method on LASFile - maybe use a lambda or put it inside the scope of the stack_curves function?

lasio/las.py
```diff
@@ -704,6 +704,43 @@ def set_data_from_df(self, df, **kwargs):
...
+        if mnemonic.isdigit():
+            return int(mnemonic)
+        else:
+            return mnemonic
+
+    def _sort_channels(self, mnemonic):
```
kinverarity1: Even more so, the logic is confusing with the name. I reckon make this a lambda too or just put it inside the scope of stack_curves?

dagrha:
Kent, thank you for the very thoughtful review! I've taken your comments to heart and reworked the whole method. One thing I did for example was change the `stub` and `curves_list` kwargs to a single arg. I think that simplifies the api, and makes it fail a little better when no argument is given.

Good suggestions to add tests for "fail cases"-- in testing for some oddball arguments (e.g. '', [''], None) I was finding that my original implementation was not failing very elegantly. This current implementation should be an improvement.

lasio/las.py
dagrha: Good point. For the moment I just intended this method to take curve mnemonics which I've now clarified in the doc string. Maybe I can add CurveItems as an option in a future PR

lasio/las.py
dagrha: Cool, glad you like it. I've maintained the functionality with my latest commit b4678cb3b4b34ff0dc19db4637b81306e7a70915 but the api changed a little

lasio/las.py
dagrha: Fair point! I decided to raise some built-in exceptions manually because the method wasn't raising "something sensible" in a couple cases where I thought it should. I hope the new implementation isn't too much.

lasio/las.py
dagrha: You are right-- silly of me to put these as methods on the LASFile object even as "private". Much cleaner to just implement as a lambda as you suggested.

kinverarity1:
Great work, thanks! I'll make a new release once merged!

kinverarity1:
Thanks @dagrha! I've made a small change to make sure the method works when `mnemonic` is a tuple or ndarray.

Issue #277, opened 2019-07-08
read does not work with pathlib.Path objects


kinverarity1:
```
Traceback (most recent call last):
  File "src\data\make_2019_geophys_log_profiles.py", line 111, in <module>
    main()
  File "c:\devapps\python\anaconda3-32\envs\gwdev2\lib\site-packages\click\core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "c:\devapps\python\anaconda3-32\envs\gwdev2\lib\site-packages\click\core.py", line 717, in main
    rv = self.invoke(ctx)
  File "c:\devapps\python\anaconda3-32\envs\gwdev2\lib\site-packages\click\core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "c:\devapps\python\anaconda3-32\envs\gwdev2\lib\site-packages\click\core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "src\data\make_2019_geophys_log_profiles.py", line 53, in main
    las = lasio.read(fn)
  File "c:\devapps\projects\lasio\lasio\__init__.py", line 41, in read
    return LASFile(file_ref, **kwargs)
  File "c:\devapps\projects\lasio\lasio\las.py", line 78, in __init__
    self.read(file_ref, **read_kwargs)
  File "c:\devapps\projects\lasio\lasio\las.py", line 113, in read
    file_obj, regexp_subs, value_null_subs, ignore_data=ignore_data, )
  File "c:\devapps\projects\lasio\lasio\reader.py", line 265, in read_file_contents
    for i, line in enumerate(file_obj):
TypeError: 'WindowsPath' object is not iterable
```

kinverarity1:
Easiest way to fix would be to check with isinstance first and if it is, convert to string.

Just have to be careful with imports, that we continue to support Python 2.6 (?) and 2.7.

kinverarity1:
Fixed by @dcslagel in #294 🎉

PR #294, opened 2019-10-25
Issue #277 read pathlib path obj

dcslagel:
### draft proposed solution for issue #277. 
 It adds a small function in reader.py to check if the file_ref instances is a descendant of the pathlib.Path object/class.  It if it is then it returns the a text stream via pathlib.Path.open().  If the file_ref is not a related to pathlib.Path it is passed through unchanged.

### python2 and python3
For this to work with python2.*, the pathlib module is added to requirements.txt.  This has a minimal impact on Python3 in that it it will install the pip pathlib module even though Python3 includes pathlib in its default modules.  

### new test
pytest -s -v tests/test_open_file.py::test_open_pathlib_object

### DIFF
--- a/lasio/reader.py
+++ b/lasio/reader.py
@@ -7,7 +7,6 @@ import traceback
 
 import numpy as np
 
-from pathlib import Path
 from . import defaults
 
 # Convoluted import for StringIO in order to support:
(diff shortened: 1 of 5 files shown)


lasio/reader.py
```diff
@@ -64,6 +71,9 @@ def open_file(file_ref, **encoding_kwargs):
         was used to decode it (if it were read from disk).
 
     """
+
+    file_ref = check_for_path_obj(file_ref)
```
dagrha: Thank you @dcslagel for giving this a go. Will be nice to have pathlib.Path support!

We may need to incorporate this into the logic following line 78. The reason is that, where it is now, it will miss all the encoding sniffing goodness that lasio performs via the `open_with_codecs` function.

dcslagel: Hi @dagrha, 

okay, change 9f45d2a adds pathlib tests to test_encoding.py to check for the the encoding of pathlib.Path objects.  

Then to make sure the Path objs go through the encoding analysis,  check_for_path_obj() now returns the full path string filename of the object.  This string should then get processed properly by 'def instance(file_ref, str):...'. 

The pathlib tests all pass and match the encodings for same las files as used in the non-pathlib tests.

See if this is a good solution or needs some additional tuning.

Thanks!,

DC

dcslagel:
Hi @dagrha, @kinverarity1,

Just following up on this proposed change and bug fix.  Let me know if this change could be accepted (or rejected) or needs some additional changes before merging.

Thanks,

DC

kinverarity1:
Hey @dcslagel, sorry, have been busy. Thanks for the reminder! I've merged with a small change, just to make sure pathlib is not a dependency.

Issue #264, opened 2019-02-05
What happens when periods are in both mnemonic and unit?


kinverarity1:
```
~CURVE INFORMATION
DEPT.FT                   :DEPTH
Speed.M/MIN                   :Speed
Cond..MS/M                   :Cond.
Gamma.CPS                   :Gamma
I. Res..OHM-M                   :I. Res.
```

Currently parsed as:

```python
[CurveItem(mnemonic=DEPT, unit=FT, value=, descr=DEPTH, original_mnemonic=DEPT, data.shape=(1590,)),
 CurveItem(mnemonic=SPEED, unit=M/MIN, value=, descr=Speed, original_mnemonic=SPEED, data.shape=(1590,)),
 CurveItem(mnemonic=COND, unit=.MS/M, value=, descr=Cond., original_mnemonic=COND, data.shape=(1590,)),
 CurveItem(mnemonic=GAMMA, unit=CPS, value=, descr=Gamma, original_mnemonic=GAMMA, data.shape=(1590,)),
 CurveItem(mnemonic=I, unit=, value=Res..OHM-M, descr=I. Res., original_mnemonic=I, data.shape=(1590,))]
```

It is very difficult to figure out which period splits the mnemonic and unit, but at the very least we should be able to exploit the long whitespace to establish that the ``value`` is wrong for the last curve.

RomCoch:
I don't know LAS1.2 specs but, according to LAS2.0 specs, line delimiters between mnemonic, units, value and description are "a) the first dot in a line, b) the first space after the first dot in a line c) the last colon in a line".
Dots shouldn't hence be used in the mnemonic...

dcslagel:
After reviewing the September 1st, 1990, LAS1.2 spec, it doesn't seem to say anything about the first space after the first dot. However the examples seem to imply that the first space after the first dot delimits the unit field.

Here is a rough-draft algorithm for the ~CURVE section (which only has API numbers and spaces in the values field). 

In the ~CURVE section    
   While the next character after the first dot is a dot    
     then throw it away and consider the next character
    -example: 
      'Cond..MS/M' would get parsed to CurveItem(mnemonic=COND, unit=MS/M,...) 

  While the next character after dot-space '. ' is not a number character or a space
    then remove it from the value field and add it to the unit field (or throw it away). 
    -example: 
      'I. Res..OHM-M' would get parsed to CurveItem(mnemonic=I, unit=Res..OHM-M, value=, ...) 

[...]

PR #295, opened 2019-12-13
Discussion and draft branch for issue 264 (What happens when periods are in both mnemonic and unit?)

dcslagel:
This branch is a discussion branch for work toward resolving issue https://github.com/kinverarity1/lasio/issues/264 (What happens when periods are in both mnemonic and unit?).  This branch is not intended for merging but rather for posting and discussing tentative solutions to the issue.

Initially it just includes a basic test for part of the issue. 

After exploring the issue a bit the main solution options seem to be:
- Add complexity to the default line parsing regular expression. (brittle?)
- Override the default regular expression for the Curve section only and use a Curve specific complex regular expression for the Curve section.
- Replace the default line parsing regular expression with parsing code.
- Change read_line(line, pattern) to read_line(line, pattern_function) and have a default and a Curve specific pattern function.

DC

### DIFF
--- a/lasio/reader.py
+++ b/lasio/reader.py
@@ -728,18 +728,40 @@ def read_header_line(line, pattern=None):
 
     """
     d = {"name": "", "unit": "", "value": "", "descr": ""}
+
+    # Default regular expressions for name, unit, value and desc fields
+    name_re = r"\.?(?P<name>[^.]*)\."
+    unit_re = r"(?P<unit>[^\s:]*)"
+    value_re = r"(?P<value>[^:]*):"
+    desc_re = r"(?P<descr>.*)"
+
+    # Alternate regular expressions for special cases
+    value_without_decs_delimiter_re = r"(?P<value>[^:]*)"
+    name_with_dots_re = r"\.?(?P<name>[^.].*[.])\."
+    no_desc_re = ""
+
+    # Configure any needed special cases
     if pattern is None:
         if not ":" in line:
-            pattern = (
-                r"\.?(?P<name>[^.]*)\." + r"(?P<unit>[^\s:]*)" + r"(?P<value>[^:]*)"
-            )
+            value_re = value_without_decs_delimiter_re
+            desc_re = no_desc_re
+            if '..' in line:
+                name_re = name_with_dots_re
         else:
-            pattern = (
-                r"\.?(?P<name>[^.]*)\."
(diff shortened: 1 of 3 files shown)


kinverarity1:
Thanks for the start on this. I think number 4 is best... I've been thinking about it, not sure what the solution is but:

- I agree complex regexps are brittle - putting logic in the code is better
- Custom parsers might help. For example my problem example in #264 is from WellCAD. It allows periods in mnemonics 😒. Instead of trying to handle it in lasio's current "universal" parser, we could try and auto-detect that it's from WellCAD, and then provide custom implementation of read_line for the Curve section (or perhaps the whole SectionParser class, though it might need improvement) to handle it. In this specific case we could exploit the fact that raw WellCAD exports have the same mnemonic and description. Of course that doesn't help when people manually edit them 😢, but we could explore that with some test cases.

The good part about custom parsers is that opens up a consistent way for providing parsers to read the service company formatted data in the ~Other section.

dcslagel:
For Discussion and feedback: 

Commit 132bca5  uses a regex approach to explore the problem.  It is based on a modification of the suggestion to try and auto-detect the WellCAD files.  It uses the '..' double dot as the symbol to check for and modify the regex pattern to accommodate the double dots.  

It makes a couple of possibly erroneous assumptions.  
1. Double dots will always be an abbreviation period followed by a the dot field delimiter.
2. WellCad las files will always use abbreviation periods on the last name along with any first name abbreviations.
(Can we make these assumptions?)

This works to pass the current test with the following command.
pytest -s -v tests/test_read.py::test_dot_delmiter_issue_264

Are there WellCad las files available for review either in the Lasio .las file examples or online?

Thanks,

DC

dcslagel:
2nd commit 132bca5, causes tests/test_write.py::test_write_sect_widths_20_wide too fail.  

I'll review and revise.

DC

dcslagel:
@kinverarity1,

3rd commit, 923dda3, changes the double-dot regex to only run in the Curve section. This both matches the issue-264 bug data and passes the current test suite.

If the abbreviations only occur in in the Curves section, limiting the mnemonics-with-period parsing to the Curves section might be a practical solution.  Are there known .las file examples with other sections containing mnemonics with periods?

Thanks,

DC

dcslagel:
I'm closing this discussion because I have posted a proposed solution:    
https://github.com/kinverarity1/lasio/pull/297

Thanks,

DC

PR #297, opened 2020-01-02
Issue #264: enable parsing periods in mnemonic field for Curve Section.

dcslagel:
Here is a proposed fix for:    
Issue #264: 'What happens when periods are in both mnemonic and unit?'

It adds a test case based on the example data in the #264 issue. All 192 tests pass on python 3.7.5.

To run the specific test:
pytest -s -v tests/test_read.py::test_dot_delmiter_issue_264 

The test file is:
tests/examples/1.2/issue-264-dot-delimiter.las

This proposed solution looks in the Curve section data for a case of 'double dots' before the description section.  If 'double dots' are found then the regular expression to parse the fields is adjusted so that the first dot is becomes a period in the mnemonic field and the second dot act as the delimiter between the mnemonic field and the unit field.

Thank you,

DC

### DIFF
--- a/tests/test_read.py
+++ b/tests/test_read.py
@@ -378,7 +378,7 @@ def test_read_incorrect_shape():
     with pytest.raises(ValueError):
         lasio.read(egfn("sample_lastcolblanked.las"))
 
-def test_dot_delmiter_issue_264():
+def test_dot_delimiter_issue_264():
     l = lasio.read(stegfn("1.2", "issue-264-dot-delimiter.las"))
     assert [c.mnemonic for c in l.curves] == [
         "DEPT",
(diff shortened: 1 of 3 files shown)


tests/test_read.py
```diff
@@ -377,3 +377,28 @@ def test_index_unit_equals_m():
 def test_read_incorrect_shape():
     with pytest.raises(ValueError):
         lasio.read(egfn("sample_lastcolblanked.las"))
+
+def test_dot_delmiter_issue_264():
```
kinverarity1: Could you correct the spelling mistake in the function name?

dcslagel: Yep, Fixing.  Thanks for catching the error.

DC

Issue #201, opened 2017-12-01
':' character in a header unit or value causes incorrect output from parser


kinverarity1:
Lines from a LAS header

```
TCS .hh:mm 21:30 23-JAN-2001:   Time Circ. Stopped
TIML.hh:mm 23:15 23-JAN-2001:   Time Logger At Bottom
```

Current behaviour - it breaks the unit/descr at the first colon from the left. 

```
HeaderItem(mnemonic=TCS, unit=hh, value=, descr=mm 21:30 23-JAN-2001:   Time Circ. Stopped, original_mnemonic=TCS)
HeaderItem(mnemonic=TIML, unit=hh, value=, descr=mm 23:15 23-JAN-2001:   Time Logger At Bottom, original_mnemonic=TIML)
``` 

Expected behavior:

```
HeaderItem(mnemonic=TCS, unit=hh:mm, value=21:30 23-JAN-2001, descr=Time Circ. Stopped, original_mnemonic=TCS)
HeaderItem(mnemonic=TIML, unit=hh:mm, value=23:15 23-JAN-2001, descr=Time Logger At Bottom, original_mnemonic=TIML)
```

kinverarity1:
I could achieve the expected behaviour in this example by breaking at the first colon reading from the right. That would introduce a new bug in the case that the descr has a colon in it, but I feel that would be less intrusive. The current behaviour breaks the unit and value (both important), and descr (not so important). The proposed solution would break only the value (important) and descr (not so important), but read the unit properly.

The ideal solution is more complex: exploit the fact that the value/descr colon is usually the same column for every line in the section, and use a likelihood function rather than a regexp to detect the value/descr colon. I don't have time to implement this at the moment.

RomCoch:
I has this problem too, specifically with with "x:long_value y:lat_value" as loc value.


> 
> 
> I could achieve the expected behaviour in this example by breaking at the first colon reading from the right. That would introduce a new bug in the case that the descr has a colon in it, but I feel that would be less intrusive. The current behaviour breaks the unit and value (both important), and descr (not so important). The proposed solution would break only the value (important) and descr (not so important), but read the unit properly.
> 
> The ideal solution is more complex: exploit the fact that the value/descr colon is usually the same column for every line in the section, and use a likelihood function rather than a regexp to detect the value/descr colon. I don't have time to implement this at the moment.

[...]

PR #298, opened 2020-01-12
Issue 201:Allow time colons before colon delimiter

dcslagel:
This change includes the change/fix for #201  and the test files to verify the change.

It incorporates test .las file examples from 
https://github.com/kinverarity1/lasio/tree/issue201-multiple-colons

And code solutions from : 
https://github.com/RomCoch/lasio/commit/71e97fdf32ea8b29c245973a047ccd45a9ce0bb2
and
https://github.com/RomCoch/lasio/commit/156def9008c385b8d3cef7c04facbafbac615496

The test cases can be run with
```
pytest -s -v tests/test_read.py::test_issue_201_non_delimiter_colon_start
pytest -s -v tests/test_read.py::test_issue_201_non_delimiter_colon_end
```

The test suite passed with this change included.

Thank you,

DC

### DIFF
--- a/lasio/reader.py
+++ b/lasio/reader.py
@@ -731,8 +731,8 @@ def read_header_line(line, pattern=None, section_name=None):
 
     # Default regular expressions for name, unit, value and desc fields
     name_re = r"\.?(?P<name>[^.]*)\."
-    unit_re = r"(?P<unit>[^\s:]*)"
-    value_re = r"(?P<value>[^:]*):"
+    unit_re = r"(?P<unit>[^\s]*)"
+    value_re = r"(?P<value>.*):"
     desc_re = r"(?P<descr>.*)"
 
     # Alternate regular expressions for special cases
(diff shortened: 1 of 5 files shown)


dcslagel:
Added failing tests. Working on fix...

DC

dcslagel:
@kinverarity1, @RomCoch,

I've added a change to fix the test failure. The change is implementing a Parameter section specific value regular expression that seeks a delimiter colon that is not surrounded by time characters. examples: 12:30, hh:mm, HH:MM.  It is limited to the 'Parameter' section for now to reduct the possibly of regression when parsing the other sections. If future testing show that this change also works for other sections it can be expanded to cover them too.

This change also adds a new test file for focused testing of specific header lines via read_header_line(). Currently it only has one test specific to Issue 201. 

Finally, it corrects the some test values in the test_read.py test_issue_201 tests.

All tests pass with this change. 

Thank you,

DC

PR #302, opened 2020-02-16
change how datetime object is turned into string so it works with latest datetime

JustinGOSSES:
I tried to use your "make a las" tutorial and found the datetime instructions didn't work. Calling str() on a datetime object fails. Instead this slightly more complicated approach is what I found others recommending when googling around. 

datetime.today().strftime('%Y-%m-%d %H:%M:%S')


kinverarity1:
Thanks!

PR #313, opened 2020-04-03
Fix Issue 182: Keep internal-las-vers when writing another version

### LINKED ISSUE #182: write(version=X) is overwriting las.version.VERS
If you call ``las.write('blah.las', version=2)`` when ``las.version.VERS.value`` is 1.2, then after the file is written, ``las.version.VERS.value`` becomes 2.0. This is incorrect behaviour. Changing the version in the write method should only affect the file being written.

The code is simple enough to change, but then it appears the wrong behaviour is happening. Needs proper  investigation and a unit test.

dcslagel:
This change will resolve #182 if accepted.

Changes:

#### 1. Add additional write(myfile, version=X) tests:

a. These test verify that the written version is different from
the internal version.

tests/test_write.py::test_write_12_to_20_ver_in_mem_is_12
tests/test_write.py::test_write_20_to_12_ver_in_mem_is_20

b. This test verifies that if the internal version is changed to 2.0, subsequently calling write() without specifying a version will write the file as 2.0 and the internal version will also be 2.0 in this case.
tests/test_write.py::test_write_12_to_20_ver

#### 2. Also added an internal comment about the verison behavior since the internal version doesn't change to the version being written to the file-stream.

Thank you,
DC

### DIFF
--- a/lasio/writer.py
+++ b/lasio/writer.py
@@ -80,17 +80,17 @@ def write(
         las.version["WRAP"] = HeaderItem("WRAP", "", "NO", "One line per depth step")
     lines = []
 
-    versionSectionToWrite = deepcopy(las.version)
+    version_section_to_write = deepcopy(las.version)
     
     assert version in (1.2, 2, None)
     if version is None:
         version = las.version["VERS"].value
     if version == 1.2:
-        versionSectionToWrite.VERS = HeaderItem(
+        version_section_to_write.VERS = HeaderItem(
             "VERS", "", 1.2, "CWLS LOG ASCII STANDARD - VERSION 1.2"
         )
     elif version == 2:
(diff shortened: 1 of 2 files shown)


lasio/writer.py
```diff
@@ -73,15 +80,17 @@ def write(
         las.version["WRAP"] = HeaderItem("WRAP", "", "NO", "One line per depth step")
     lines = []
 
+    versionSectionToWrite = deepcopy(las.version)
```
kinverarity1, line 83: Change camel case to underscore separated lower case, to keep styles consistent

dcslagel: Commit d427d2c, changes the versionSectionToWrite variable name  to version_section_to_write. 

Thanks,
DC

Issue #310, opened 2020-03-28
LASFile.metadata is broken?


kinverarity1:
```python
>>> import lasio.examples
>>> las = lasio.examples.open('sample.las')
>>> las.metadata
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\kinve\code\lasio\lasio\las.py", line 602, in metadata
    s.append(item)
  File "C:\Users\kinve\code\lasio\lasio\las_items.py", line 410, in append
    self.assign_duplicate_suffixes(newitem.useful_mnemonic)
AttributeError: 'str' object has no attribute 'useful_mnemonic'
```

dcslagel:
**Triage notes:**

1. first note
It looks like LASFile.metadata is sending LASFile.section names to SectionItems.append().  But SectionItems.append() is currently expecting dictionary with a useful_mnemonic key/value.

Here is the code from LASFile.metadata()
```python
    def metadata(self):
        s = SectionItems()
        for section in self.sections:
            for item in section:
                s.append(item)
        return s
```

It looks like, at least in some cases, 'section' is a string rather than a dictionary object and subsequently 'item' is a single letter from the string.  

2. Second Node 
A working version of passing items to SectionItems.append() is the parse_header_section() function in lasio/reader.py.   It is doing a quite different set of actions but there may be a few specific concepts that should be implemented in LASFile.metadata() to fix the current break

3. Third Note
I ran the sequence of steps that Kent described in the initial error report on various tags 0.8 , v0.12, v0.14, v0.16 and v0.25.  This issue exists on all those tags,  this issue probably always existed and probably not a regression.

DC

JustinGOSSES:
Found this issue when trying to work on notebook using lasio, wellio, and wellioviz together https://github.com/JustinGOSSES/wellio.js/pull/48

When working with multiple logs, wellioviz uses well name to organize them. Lasio's generated JSON drops well header information such as well name.

kinverarity1:
Thanks @justinGOSSES I'll have a look.
________________________________

kinverarity1:
I'll continue the discussion at wellio: https://github.com/JustinGOSSES/wellio.js/pull/48#issuecomment-612327542

kinverarity1:
Fix was released in v0.25.1

Issue #312, opened 2020-04-03
Test Failure: test_api.py::test_version failing 


dcslagel:
test_api.py::test_version fails after commit  093fd03 'add setuptools_scm version code'

Failure message:
```python
    def test_version():
>       assert '.' in lasio.__version__
E       AttributeError: module 'lasio' has no attribute '__version__'


```

It looks like the issue is related to a change in lasio/__init__.py:

```diff
+from pkg_resources import get_distribution, DistributionNotFound
+
+try:
+    __version__ = get_distribution(__name__).version
+except DistributionNotFound:
+    # package is not installed
+    pass
+
 import os
 
 from .las import LASFile, JSONEncoder
@@ -12,7 +20,7 @@ else:
     from .excel import ExcelConverter
 
 
-__version__ = '0.25.0'
+
```

PR #314, opened 2020-04-05
Issue-310: Implement draft solution to fix LASFile.metadata

dcslagel:
This branch explores possible solutions to issue #310 

This commit explores two changes.
1. Cycle through self.sections.items() instead of self.sections so we can access both the section_names and SectionItem objects.

2. Check that each SectionItem object has an
assign_duplicate_suffixes() method. If the object doesn't have this method, then log a warning and continue to on to the next section.
This will occur on a SectionItem that includes a header item which isn't parsed into a HeaderItem object (mnemonic, unit, value, desc).  An example of this an the '~Other' section that parsed into a str object and only contains a text comment (tests/example/sample.las).

Here is the SectionItem.append() code that a non-HeaderItem section will fail on if it doesn't have 
```python
    SectionItem.append(self, newitem):
        '''Append a new HeaderItem to the object.'''
        super(SectionItems, self).append(newitem)
        self.assign_duplicate_suffixes(newitem.useful_mnemonic)
```

These changes enable LASFile.metadata to display the metadata for all header information that is in the HeaderItem object format.  Should additional changes be made to enable handling string object metadata/headers or other non-standard header information?  

Thanks,
DC

### DIFF
--- a/lasio/las.py
+++ b/lasio/las.py
@@ -589,32 +589,6 @@ class LASFile(object):
         self.sections["Other"] = section
 
     @property
-    def metadata(self):
-        """All header information joined together.
-
-        Returns:
-            :class:`lasio.SectionItems` object.
-
-        """
-        s = SectionItems()
-        for section_name, section in self.sections.items():
-            if not hasattr(section, 'assign_duplicate_suffixes'):
-                logger.warning( """
-WARNING: Section skipped: It doesn't have an assign_duplicate_suffixes method.
-    SectionName: %s
-    SectionContent:
-        %s\n""" % (section_name, section)
-                )
-                continue
-            for item in section:
-                s.append(item)
-        return s
-
-    @metadata.setter
-    def metadata(self, value):
-        raise NotImplementedError("Set values in the section directly")
-
-    @property
     def header(self):
         """All header information
 
--- a/lasio/las.py
+++ b/lasio/las.py
@@ -597,7 +597,15 @@ class LASFile(object):
 
         """
         s = SectionItems()
-        for section in self.sections:
+        for section_name, section in self.sections.items():
+            if not hasattr(section, 'assign_duplicate_suffixes'):
+                logger.warning( """
+WARNING: Section skipped: It doesn't have an assign_duplicate_suffixes method.
+    SectionName: %s
+    SectionContent:
+        %s\n""" % (section_name, section)
+                )
+                continue
             for item in section:
                 s.append(item)
         return s


dcslagel:
Further thoughts:  '~Other' and other non-standard header items probably should have a way to be processed as SectionItems and displayed with LASFile.metadata.   Looking further at that.

DC

kinverarity1:
Thanks @dcslagel! Should we drop the attribute altogether? With duplicate mnemonics and potentially even duplicate sections once we have LAS 3 (i.e. multiple runs with repeated ~P sections)  it might be limited in usefulness?

dcslagel:
@kinverarity1,
Do you mean drop (remove) the 'metadata' attribute from LASFile ?   If so, that occurred to me as well since it looks like LASFile.metadata hasn't worked for a many versions. As I mentioned on ticket #310,  I ran the sequence of steps that you described in the initial error report on various tags 0.8 , v0.12, v0.14, v0.16 and v0.25. This issue exists on all those tags, this issue probably always existed and probably not a regression.

However I hesitated to suggest that, because I figured there was a use for the LASFile.metadata attribute.  I'll attempt a separate branch with LASFile.metadata removed and post it for a pull-request, then you can make the decision on it.

On a related note, should SectionsItems objects be able to handle '~Other' and other non-standard header items regardless of the existence of LASFile.metadata?

Thanks!,
DC

kinverarity1:
@dcslagel Yep, I was thinking, maybe we should just remove LASFile.metadata altogether. Sorry, I should have mentioned this on the issue.

About ~Other - I'm not sure, happy to be guided by what you think is do-able. It makes sense that all the LAS file sections should have the same type, but are we still able to keep a simple interface such that it is still possible to assign a string to the section e.g. (pseudocode)

```python
>>> las.other = "Blah blah"
>>> type(las.other)
<SectionItems>
```

dcslagel:
Commit 1a6d958 removes LASFile.metadata  property and methods.  Everything in the test suite still passes except pre-existing failure 'test_version' test (noted in #312).

Additionally I checked the documentation and didn't see anything specific to  LASFile.metadata so no changes there.  I did a check of the notebook files by greping for "metadata " and didn't find anything so we should be okay there.

Side note:  I spent some time looking into how to make the ~Other SectionItem more behave the same as other 'Item's in a SectinItem list.  It would be a significant refactoring and at this time I don't see a significant use benefit.  If it comes up in the future as a useful change, I'll be interested to look at it at that time. 

Thank you,

DC

PR #316, opened 2020-04-16
WIP: Issue-312: Add a reasonable non-dist lasio.__version__ option

dcslagel:
@kinverarity1,

This is a proposed solution for #312 "Test Failure: test_api.py::test_version failing".  

Its solution is to set a sensible version string when lasio is installed from GitHub and doesn't have a distribution tag-based version string.   It is in 'draft'/'work-in-progress' mode because I still need to update the version test for this case.

Let me know this will be a satisfactory solution or if there are changes/improvements I can make  .

Thank you,
DC

### DIFF
--- a/lasio/__init__.py
+++ b/lasio/__init__.py
@@ -1,9 +1,17 @@
+import subprocess
 from pkg_resources import get_distribution, DistributionNotFound
 
 try:
     __version__ = get_distribution(__name__).version
 except DistributionNotFound:
-    # package is not installed
+    __version__ = "A Dev version: probably in a git repo."
+    tmpbytes = subprocess.check_output(
+        ["git", "log", "-1", "--pretty=tformat:Commit %h"]
+    ).strip()
+
+    tmpstr = "".join( chr(x) for x in tmpbytes)
+    if tmpstr.startswith("Commit"):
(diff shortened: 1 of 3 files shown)


kinverarity1:
Thanks for picking this up! I'm happy with this approach, I don't know much about it. I will merge once you are ready.

dcslagel:
Commit e570acd changes to using 'python setup.py --version' to obtain a version string when the current lasio installation is not from a pip install or distribution package (zip, rpm, etc).  
This works for git repo clones, but fails with a Traceback if the install is not a git repo.  

Continuing to look for a more resilient set of version fall-backs....

Dc

dcslagel:
Commit c93df45 has the following changes:

1. Move version functions to a separate file: 'las_version.py.
2. Look for a version string first if in a release distribution package.
3. If not a distribution try to set the version string via git
4. If not in a Lasio git repo, set a sensible default version based on 0.25.0. 0.25.0 is the last release before setup tools_scm was introduced and the last release before this change.

Still need to update the las_version.py comments.
DC

dcslagel:
@kinverarity1,

I've made the cleaned-up change as Pull-Request #318 and  closing exploratory draft pull request.

Thanks,
DC

PR #318, opened 2020-04-23
Issue 312: Update lasio.__version__ for various situations

### LINKED ISSUE #312: Test Failure: test_api.py::test_version failing

dcslagel:
This change will resolve #312 if accepted.

- Move version code to separate file, keeping lasio.__version__.
- Add lasio.version() to fix the entry point "lasio = lasio:version"
- Add attempts to identify a reasonable version string for various
  environment situations: official package, local package, editiable
  install, git-repo without an install, and a plain diectory.
- Verify that test_api.py::test_version passes in each environment
  situation.

Thank you,
DC

### DIFF
--- a/lasio/__init__.py
+++ b/lasio/__init__.py
@@ -1,16 +1,12 @@
-from pkg_resources import get_distribution, DistributionNotFound
-
-try:
-    __version__ = get_distribution(__name__).version
-except DistributionNotFound:
-    # package is not installed
-    pass
 
 import os
 
+from .las_version import version
 from .las import LASFile, JSONEncoder
 from .las_items import CurveItem, HeaderItem, SectionItems
 from .reader import open_file
+
+__version__ = version()
 
 try:
     import openpyxl
--- a/lasio/las_version.py
+++ b/lasio/las_version.py
@@ -1,0 +1,91 @@
+import subprocess, re
+from pkg_resources import get_distribution, DistributionNotFound
+from datetime import datetime
+
+# d[year][month][day] example: 20200420
+ver_date = datetime.now().strftime("d%Y%m%d")
+
+def version():
+    las_version = ''
+
+    '''
+    Look for distribution version
+
+    This looks for a distribution in roughly the following order.
+    - Local development install installed with 'pip install -e "."'
+        Note:
+            The '-e' puts the install in an editable-mode where it it looks at
+            the source directory and reactes to changes in the code.
+    - A lasio.egg-info dir in the current working directory.
+        Note!:
+            An existing lasio.egg-info dir can hide an offical release install
+            for the case where an offical release is also installed.  The
+            lasio.egg-info can be an artifact of running 'pip install -e "."',
+            or of runnng 'bdist_wheel' to create a release package.
+    - An official release installed with 'pip install lasio' or a development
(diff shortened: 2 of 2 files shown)


kinverarity1:
Hi @dcslagel. Thanks for all your work on resolving the issues with my change to setuptools_scm. I didn't realise it would be so tricky to cover all the edge cases.

What would you recommend is the next best step - should I merge both #317 and #318 into master, then create a new release with ``git tag v0.25.1``?

dcslagel:
Hi @kinverarity1,

**Main thought:**

I (strongly) feel that it would be best to create a release with `git tag v0.25.1` BEFORE merging #317 and #318. 

The thinking is that this will be the first release with setuptools_scm and we want to check for changes in the release steps and any unexpected issues in only the new setuptools_scm changes before adding any additional code.

**Additional thoughts:**

Yes!, it is surprising how many different version cases there are. Fortunately, most users will be doing `pip install lasio`, and get the expected release version tag/string.  

A smaller set of folks editing the lasio-repo code will get the more complicated cases.  For most of the cases, the new setuptools_scm versioning with provide a more accurate version that includes the commits since the release-tag that is the base. So that is a nice addition! :-)

There are a few situations where a developer may not be sure how the version is being derived. In those cases, reading las_version.py should clarify where the current version is being derived from.

Thanks,
DC

kinverarity1:
I've released v0.25.1... it seemed to run smoothly? Let me know if you think these PRs should be merged.

Thanks heaps for doing these. I'm learning a lot!

dcslagel:
Hi @kinverarity1,

Yes, these PRs can be merged.  
I pulled in v0.25.1; then, in various ways including the pip install, egg/wheel installs and source import and re-checked the various lasio. output.

They are the expected version strings as follows:
|context|reported version|
|-------|-----------------|
|'pip install lasio'| 0.25.1|
|master+my-merge-branch (source import)| 0.25.1.dev4+g25c3e62.d20200501| 
|'pip install -e' (create egg and install)| 0.25.**2**.dev4+g25c3e62.d20200501|
|non-git-source dir (source import)| 0.25.0.dev0+unknown-post-dist-version.d20200501|

And with running the entry-point-lasio on the command line `$ lasio`
|context|reported version|
|-------|-----------------|
|'pip install lasio'| cannot import name 'version'..|
|'pip install .' (create egg and install)| 0.25.**2**.dev4+g25c3e62.d20200501|
|no install | lasio; No such file or directory|

Also I checked that pip install lasio[test] will install the packages needed for running the test suite.

Thank you,
DC

PR #334, opened 2020-05-11
Fix deprecation warnings reported in test runs

dcslagel:
- test_read.py::test_missing_a_section
  DeprecationWarning: use `array.size > 0` to check that an array is
  not empty. In this test we check that the array is empty, so `not
  array.size > 0`.
- las/las_version.py
  DeprecationWarning: invalid escape sequence \d
  The solution is to use raw strings for regexs: r"\d" instead of "\d".

Thank you,
DC

### DIFF
--- a/lasio/las_version.py
+++ b/lasio/las_version.py
@@ -54,7 +54,7 @@ def version():
 
 
 def _get_vcs_version():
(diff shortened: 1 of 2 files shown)


kinverarity1:
Thank you for addressing these!

Issue #333, opened 2020-05-11
Header item parsed incorrectly when colon is in description


kinverarity1:
This example LAS file recently added to the lasio issue tracker has a header item where the description is parsed incorrectly:

```
QI     .      :         Survey quality: GOOD or BAD versus criteria
```

Currently the result is:

```
{'name': 'QI',
 'unit': '',
 'value': ':         Survey quality',
 'descr': 'GOOD or BAD versus criteria'}
```

The expected output is:

```
{'name': 'QI',
 'unit': '',
 'value': '',
 'descr': 'Survey quality: GOOD or BAD versus criteria'}
```

kinverarity1:
The fix for this issue was released in [v0.26](https://github.com/kinverarity1/lasio/releases/tag/v0.26)

PR #335, opened 2020-05-11
Add fix for meta-data parameter section parsing for #333

dcslagel:
This should fix https://github.com/kinverarity1/lasio/issues/333.
It is also related to https://github.com/agile-geoscience/welly/issues/128

- Change the parameter-section's value parsing to first attempt to catch
  values that might have a time-string in them and then if no value is
  found try the standard value_re regular expression. This combination
  is handling all the current test cases.
- Move pattern configuration task to a separate function because it is
  continuing to expand and be complex enough to warrant its own
  function.

Thank you,
DC

### DIFF
--- a/lasio/reader.py
+++ b/lasio/reader.py
@@ -324,7 +324,7 @@ def read_file_contents(file_obj, regexp_subs, value_null_subs, ignore_data=False
                }
                sect_lines = []
                sect_line_nos = []
-        
+
 
             if not ignore_data:
                 try:
@@ -743,6 +743,9 @@ def read_header_line(line, pattern=None, section_name=None):
 
     Arguments:
         line (str): line from a LAS header section
+        section_name (str): Name of the section the 'line' is from. The default
+        value is None.
+
 
     Returns:
         A dictionary with keys 'name', 'unit', 'value', and 'descr', each
@@ -756,18 +759,16 @@ def read_header_line(line, pattern=None, section_name=None):
     m = None
 
     if pattern is None:
-        patterns = configure_and_match_pattern(line, section_name)
+        patterns = configure_metadata_patterns(line, section_name)
     else: # pattern was passed in on function call
         patterns.append(pattern)
-        m = re.match(pattern, line)
 
     for pattern in patterns:
+        # Attempt to parse the section line's name(mnemonic), unit, value and
+        # descr fields with the given pattern.
         m = re.match(pattern, line)
         if m is not None:
             break
-
-    if m is None:
-        logger.warning("Unable to parse line as LAS header: {}".format(line))
 
     mdict = m.groupdict()
     for key, value in mdict.items():
@@ -777,18 +778,17 @@ def read_header_line(line, pattern=None, section_name=None):
                 d[key] = d[key].strip(".")  # see issue #36
     return d
 
-def configure_and_match_pattern(line, section_name):
-    """Configure a regular expression pattern and try to match it with the line
+def configure_metadata_patterns(line, section_name):
+    """Configure regular-expression patterns to parse section meta-data lines.
 
     Arguments:
         line (str): line from LAS header section
-        section_name (str): Name of the section the 'line' is from
+        section_name (str): Name of the section the 'line' is from.
 
     Returns:
-        A dictionary with keys 'name', 'unit', 'value', and 'descr', each
-        containing a string as value or the None value.
-
+        An array of regular-expression strings (patterns).
     """
+
     # Default return value
     patterns = []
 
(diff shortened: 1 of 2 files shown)


lasio/reader.py
```diff
@@ -751,6 +751,47 @@ def read_header_line(line, pattern=None, section_name=None):
     """
```
kinverarity1, line 751: It would help to have a mention of the `pattern` and `section_name` keyword arguments by referring the reader to the `configure_and_match_pattern()` docstring instead.

lasio/reader.py
```diff
@@ -751,6 +751,47 @@ def read_header_line(line, pattern=None, section_name=None):
...
 
+    # Set defaults for local variables.
+    patterns = []
+    m = None
+
+    if pattern is None:
+        patterns = configure_and_match_pattern(line, section_name)
+    else: # pattern was passed in on function call
+        patterns.append(pattern)
+        m = re.match(pattern, line)
```
kinverarity1, line 762: Is line 762 superfluous? The result in `m` is presumably overwritten at line 765? If my understanding is correct it would be better to remove line 762.

lasio/reader.py
```diff
@@ -751,6 +751,47 @@ def read_header_line(line, pattern=None, section_name=None):
...
+
+    for pattern in patterns:
+        m = re.match(pattern, line)
+        if m is not None:
+            break
+
+    if m is None:
+        logger.warning("Unable to parse line as LAS header: {}".format(line))
```
kinverarity1, line 770: To clarify something: does this mean that when we have a malformed header line the following will happen:

If ignore_header_errors is False:
1. a logger.warning message is emitted here
2. an exception occurs here at line 772
3. In `read_header_section()`, line 559 the exception is caught an re-raised as a LASHeaderError
4. parsing stops

If ignore_header_errors is True:
1. a logger.warning message is emitted here
2. a logger.warning message is again emitted at line 557 in `read_header_section()`
3. parsing continues

If I'm correct, I think we should remove the logger.warning call here and accept that because an exception is raised there is no need for a logger.warning call. What do you think?

lasio/reader.py
dcslagel: Yep, it is superfluous, good catch. fixing in next commit.

lasio/reader.py
dcslagel: I'm adding section_name and pattern argument information to read_header_line() docstring so there is no-jumping around looking for the information.  Also the Return value description on configure_and_match_pattern() will be updated to state that it is an array of regular expression strings (patterns).

lasio/reader.py
dcslagel: Yes, That is right.  Removing the line 770 logger.warning in the next commit.  

One side note for ignore_headers_errors=False:
The exception occurring at line 772, first reports an `AttributeError: 'NoneType.. During Handling another exception occurred: ...` and then `LASHeaderError..`. So 2 exceptions will be displayed and the root cause is at the first one.  I think this is okay, but let me know if it should be changed.

kinverarity1: Good point! It would be better to have just the one exception raised from read_header_line(), I didn't really think too closely about it. That would be a good idea, if you're willing to rearrange it, but it's not necessary, I don't mind either way.

dcslagel: I haven't found the way yet to arrange things to only raise the LASHeaderError and have it include the relevant AttributeError information.  So if you don't mind, I'll work it out in a future pull-request.

kinverarity1: That's totally fine! Thanks so much for all this!

dcslagel: @kinverarity1,

I spent time today to look at ways of simplifying or combining the two exceptions `LASHeaderError` and `AttributeError` in  parse_header_section().  The outcome is that so far having both display is better than other solutions I found.  So I think we should leave them be unless some future information provides a better way.

Details:
- I could make one display and not the other but in those cases some useful debugging information was lost.  If only AttributeError  is displayed then the real line number of the error is in the stack trace but the line of las meta data being acted on is dropped.  If only LASHeaderError is displayed then the real line number of the error is dropped.

- There are way to programmatically combine the two exceptions but that added specific and unusual code to the exception handling and cluttered the regular flow of Lasio code with exception specific code.  So that doesn't seem like a good approach.

Given these finding, I won't make a PR for changing these two exceptions at this time. 

Let me know if you have additional thoughts on it.

Thanks!,
DC

kinverarity1:
The code in this PR was released as part of [v0.26](https://github.com/kinverarity1/lasio/releases/tag/v0.26)

PR #340, opened 2020-05-22
Recognise 0.1 inch as an index unit

kinverarity1:
This draft PR would recognise `.1IN` as an index unit and provide automatic conversions to feet and metres - fixing #339.


DaveHealy-Aberdeen:
Yes, this works for me - and those numbers in the example look correct. 
Thanks
Dave

kinverarity1:
Thanks Dave, I just need to add a regression test for this and will then merge it.

kinverarity1:
Thanks @dcslagel this is now done!

PR #342, opened 2020-05-24
Remove unused imports

MandarJKulkarni:


lasio/reader.py
```diff
@@ -2,12 +2,10 @@
 import logging
 import os
 import re
-import textwrap
 import traceback
 
 import numpy as np
 
-from . import defaults
```
kinverarity1, line 10: This is used in reader.py:

- https://github.com/kinverarity1/lasio/blob/b001ae1d92ac2afdaa77787b0d3110a82fc60ac3/lasio/reader.py#L464-L465
- https://github.com/kinverarity1/lasio/blob/b001ae1d92ac2afdaa77787b0d3110a82fc60ac3/lasio/reader.py#L600
- https://github.com/kinverarity1/lasio/blob/b001ae1d92ac2afdaa77787b0d3110a82fc60ac3/lasio/reader.py#L639

kinverarity1:
sorry internet hiccup - accidentally sent the review twice.

lasio/reader.py
MandarJKulkarni: Yes, but it was imported twice. Once at line 10 and then at line 30. I had removed the first import.
Not sure why Python2.7 build is failing, it is not starting the tests. 
But failing for command "python setup.py egg_info:" after taking ~8 mins.

kinverarity1: Oh whoops - thanks for picking that up, we can definitely remove the duplicate line in that case. Sorry.

kinverarity1:
I'll investigate the python 2.7 test failure... Can't understand why it's happening with this pr only

dcslagel:
I took downloaded MandarJKulkarni:patch-1 and installed python 2.7.15 (same version as the failed build is running) on a local machine.  Then ran the configured commands for the failing Travis-CI job:
```python
 "install": [
    "pip install --editable \".[all]\"",
    "pip install pytest>=3.6 pytest-cov coverage pathlib"
  ],
  "script": [
    "pytest"
  ],
```

Everything worked and the tests all passed.

So the pull-request seems fine.  Maybe this Travis-CI instance has some issue.   

It might be worthwhile to copy the branch and create a new pull-request to see if Travis-CI passes.

DC

kinverarity1:
Thanks DC

kinverarity1:
You were right #344 passed Travis fine so it was just some glitch. Sorry about the hold up in merging this @MandarJKulkarni  - thanks for the PR

PR #346, opened 2020-05-27
Master Branch: Base leading zero test files on sample.las

dcslagel:
Change the leading zero test files to be sample*.las plus the leading
zero special case.  Previously the leading zero test files had other
special cases in them that caused their tests to exercise unrelated code
paths.

Thank you,
DC


kinverarity1:
Thanks 😀

dcslagel:
Thank you for accepting this change. :-)

PR #349, opened 2020-05-28
Create test for 0.1 inch index unit: issue-339

dcslagel:
This pull request is to add a test case for the branch issue339-tenthinch.

- pytest tests/test_enhancements.py::test_autodepthindex_point_one_inch
- test that index unit is .1IN
- test that depth_m[...] will convert to .1IN index
- test that depth_ft[...] will convert to .1IN index

Thank you,
DC


kinverarity1:
Thanks!

Issue #350, opened 2020-05-29
test_mnemonic_case_comparison_lower sets well.null to "" instead of -999.2500


dcslagel:
The test test_mnemonic_case_comparison_lower calls the 1.2 file: mnemonic_case.las.
Around line 203  self.well.null.value is assigned to the variable 'null' but self.well.null.value is an empty string while self.well.null.descr has the actual value -999.2500.

Other 1.2 las files do not have this bug. A run of pytest tests/test_read.py::test_read_v12_sample, successfully sets the variable 'null' to -999.25. 

version: master branch at 096790b.

The issue can be seen by running the test case on the work-in-progress branch at:
https://github.com/kinverarity1/lasio/pull/351

DC

kinverarity1:
Nasty! Thank you for catching that!

dcslagel:
Thanks, :-)   

Yeah, it was interesting,  It was one of the test cases reported in a FutureWarning at the end of a pytest run.
```
...
tests/test_sectionitems.py::test_mnemonic_case_comparison_lower
FutureWarning: elementwise comparison failed; 
returning scalar instead, but in the future will perform elementwise comparison
    arr[arr == null] = np.nan
```

While trying to figure out the issue, I noticed that `null` was set to an empty string ''.  From there, it lead to the lowercase values not getting set!

DC

PR #351, opened 2020-05-29
Add test case for well null value

dcslagel:
This is a working branch for #350. 

The failing test case is here:
pytest -s -vv tests/test_sectionitems.py::test_mnemonic_case_comparison_lower

Output:
```python
    def test_mnemonic_case_comparison_lower():
        las = lasio.read(egfn('mnemonic_case.las'), mnemonic_case='lower')
        assert 'DEPT' in las.curves
>       assert las.well.null.value == -999.25
E       AssertionError: assert '' == -999.25
E         +''
E         --999.25

tests/test_sectionitems.py:130: AssertionError
```


dcslagel:
This draft-pull-request is replace by the pull-request #352.

PR #353, opened 2020-06-01
To rearrange-reader branch: add basic las 30 test

dcslagel:
LAS 3.0

This pull request is to get a standard LAS 3.0 test file on the rearrange-reader branch so we can exercise the new reader with a LAS 3.0 file.

- Add tests/examples/3.0 dir.
- Add the CWLS's 3.0 example las file.
- Copy the example file to sample_3.0.las to standardize with
  1.2 and 2.0 sample las files.
- Create a tests/test_read_30.py with basic read test.
  However, the test is set to SKIP because it currently fails on the rearrange-reader branch.

Thank you,
DC


kinverarity1:
Thanks, that's great!

PR #356, opened 2020-06-02
LAS 3.0 : Rearrange reader branch: find sections revision : Work in progress

dcslagel:
This is a work-in-progress pull request to look at the issue of finding section headers for LAS 3.0

I have a doubt that we can combine text based readline()/enumerate of a file with binary based file.seek()s given that LAS files can be in different encodings and contain multi-byte characters.  

If this is correct then we can get the section locations ( '~' followed by b"\n") by opening the file in binary mode, gathering the locations, then closing the file when done.   This is what this branch is looking at.

Another option is the read the file in text mode and extract the sections as we go.  This would mean holding more of the data in memory.  Is that a consideration?

This issue came visible when working to resolve the tests/test_encoding.py test failures:
`pytest -s -vv tests/test_encoding.py::test_encoding_attr`
which outputs
```
KeyError: 'Unknown section name 00091604920W300' 
```

DC


kinverarity1:
Ah, right. Of course! I was initially stumped by the encoding errors, but this explains it. I'm not sure how to solve it but I'd prefer not to hold the file in memory as we had some similar issues previously (#170).

dcslagel:
04cb355   fixes test_encoding_attr and some other test_encoding.py tests.
    
However, it breaks a bunch of other tests, particularly in test_read.py.
Worse, it seems to hang when running test_enhancements.py. It could be terrible performance or a loop.
    
So this is just a checkpoint commit, not a resilient solution.

dcslagel:
Closing this exploratory branch.  It is replace with the active pull request at:    
https://github.com/kinverarity1/lasio/pull/357  LAS 3.0 : Rearrange reader branch: Change find_sections to use byte string

PR #357, opened 2020-06-04
LAS 3.0 : Rearrange reader branch: Change find_sections to use byte string

dcslagel:
This pull-request is targeted for the rearrange-reader branch. It changes reader.py::find_sections_in_file().

- The find_sections_in_file() interface is changed to take a file_ref instead of file_obj.
- The return object has the same format except it returns a binary string instead of string for each section header.
- In las.py decode section_title from binary string to section_title_str for the expected functionality.

After this change all tests run with reasonable speed for development purposes.
29 test fail as follows:
```
FAILED tests/test_encoding.py::test_utf16lebom_cchardet
FAILED tests/test_encoding.py::test_utf16le_cchardet_fails
FAILED tests/test_encoding.py::test_utf16bebom_cchardet
FAILED tests/test_encoding.py::test_pathlib_utf16lebom_cchardet
FAILED tests/test_encoding.py::test_pathlib_utf16le_cchardet_fails
FAILED tests/test_encoding.py::test_pathlib_utf16bebom_cchardet
FAILED tests/test_examples.py::test_github
FAILED tests/test_json.py::test_json_null
FAILED tests/test_null_policy.py::test_null_policy_NULL_none
FAILED tests/test_null_policy.py::test_null_policy_custom_2
FAILED tests/test_null_policy.py::test_null_policy_runon_replaced_1
FAILED tests/test_null_policy.py::test_null_policy_runon_replaced_2
FAILED tests/test_null_policy.py::test_null_policy_runon_ok_1
FAILED tests/test_null_policy.py::test_null_policy_runon_ok_2
FAILED tests/test_open_file.py::test_open_url
FAILED tests/test_open_file.py::test_open_file_object
FAILED tests/test_open_file.py::test_open_string
FAILED tests/test_write.py::test_write_sect_widths_12
FAILED tests/test_write.py::test_write_sect_widths_20_narrow
FAILED tests/test_write.py::test_write_sect_widths_20_wide
FAILED tests/test_write.py::test_df_curve_addition_on_export
FAILED tests/test_write.py::test_multi_curve_mnemonics_rewrite
FAILED tests/test_write.py::test_multi_curve_missing_mnemonics_rewrite
FAILED tests/test_write.py::test_write_units
FAILED tests/test_write.py::test_rename_and_write_curve_mnemonic
FAILED tests/test_write.py::test_write_single_step
FAILED tests/test_write.py::test_write_12_to_20_ver_in_mem_is_12
[...]

PR #358, opened 2020-06-05
LAS 3.0 : Rearrange reader branch: Fix find_sections_in_file with text stream

dcslagel:
This pull request is an alternative to https://github.com/kinverarity1/lasio/pull/357.

This request retains most of the readline() approach with changes to get the correct section positions.  It also replaces codecs.open() with the more current io.open.  However the other codecs encoding logic is still there.

It passes the test suite on Python 3.7  with 21 tests failing

Currently it fails to complete the test suite on Python 2.7 with the error: 
```UnicodeEncodeError: 'ascii' codec can't encode characters```
This can be worked around.

Is either approach more preferred than the other?

Thanks!

DC

### DIFF
--- a/lasio/reader.py
+++ b/lasio/reader.py
@@ -1,4 +1,5 @@
 import codecs
+import io
 import logging
 import os
 import re
@@ -135,7 +136,7 @@ def open_with_codecs(
 
     Keyword Arguments:
         encoding (str): character encoding to open file_ref with, using
-            :func:`codecs.open`.
+            :func:`io.open`.
         encoding_errors (str): 'strict', 'replace' (default), 'ignore' - how to
             handle errors with encodings (see
             `this section
@@ -193,7 +194,7 @@ def open_with_codecs(
             filename, encoding, encoding_errors
         )
     )
-    file_obj = codecs.open(
+    file_obj = io.open(
         filename, mode="r", encoding=encoding, errors=encoding_errors
     )
     return file_obj, encoding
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by kinverarity1** [comment]
Thanks for both of these! I will have a closer look this weekend...

**Comment by dcslagel** [comment]
Great.  Thanks, I appreciate your taking the time to look these over.

File: lasio/reader.py [hunk 1]
```diff
@@ -270,27 +271,33 @@ def find_sections_in_file(file_obj):
...
 
     """
-    k = 0
+    file_pos = int(file_obj.tell())
     starts = []
     ends = []
-    for i, line in enumerate(file_obj):
+    line_no = 0
+    line = file_obj.readline()
+    # for i, line in enumerate(file_obj):
```
dcslagel, line 286: Note: io.TextIOWrapper 'locks' file_obj.tell() if run in a for loop but doesn't in the while statement.  That is why the change from ```for``` to ```while```.

lasio/reader.py
```diff
@@ -270,27 +271,33 @@ def find_sections_in_file(file_obj):
...
+    while line:
         sline = line.strip().strip("\n")
         if sline.startswith("~"):
-            starts.append((k, i, sline))
+            starts.append((file_pos, line_no, sline))
```
dcslagel, line 290: Changing ```k``` and ```i``` to descriptive names ```file_pos``` and ```line_no``` just helped me keep the logic in my head.  Let me know if ```k``` and ```i``` are preferred.

dcslagel:
Yes, I agree with you, #358 is clearer and will be more manageable to maintain.  Thank you for making the time to review these!

Regarding the issue in the Python 2.7 tests `UnicodeEncodeError: 'ascii' codec can't encode characters`,  what are the Lasio plans for Python 2.7 support?  How long will Lasio continue support for 2.7? 

It looks like the Python Foundation ended support on January 1, this year.
https://www.python.org/doc/sunset-python-2/

Thanks,
DC

kinverarity1:
We can drop support for 2.7 as soon as we have to... It's time 😎

Issue #373, opened 2020-07-23
Add github action with testing


dbhart:
Would provide github CI testing of the code, allowing for a badge to be placed showing the tests are passing.

dbhart:
Specifically, this is referencing the ability to have multiple OSes, which travis-ci does not provide.

kinverarity1:
#374 and #387 have been merged to master and GitHub CI testing is now happening. Thank you @dbhart!

PR #374, opened 2020-07-23
GitHub action added for automatic testing

dbhart:
Proposed github action to perform testing on push and pull_requests. See issue #373


.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install numpy cchardet openpyxl pandas
```
kinverarity1, line 26: Shall we stick to the existing setup commands in the Travis CI script?

https://github.com/kinverarity1/lasio/blob/bd42e823653966472c2e2cca29163ed3e5a602f3/.travis.yml#L12-L13

.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,30 @@
...
+        pip install pytest coverage pytest-cov codecov pathlib
+    - name: Test with pytest
+      run: |
+        pytest --cov-fail-under=80 --cov-branch
```
kinverarity1, line 30: We should keep the test command as `pytest`. Any command-line arguments can go into  [pytest.ini](https://github.com/kinverarity1/lasio/blob/master/pytest.ini).

What do `--cov-fail-under=80` and `--cov-branch` do - are they needed?

.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,30 @@
...
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        python-version: [3.5, 3.6, 3.7, 3.8]
+        os: [ubuntu-latest, windows-latest, macOS-latest]
```
kinverarity1, line 15: :heart: Love it!

dcslagel:
@kinverarity1 @dbhart,  Just now re-looking at this. 

 I'm not yet familiar with GitHub-Actions. Is the main benefit of moving to GitHub Actions is that it will run tests on linux, windows and mac?  If that is correct, then that is worthwhile.  

I'll exercise this a bit to understand better and have some additional comments in a few days.

It looks like @dbhart has some good test results at:
https://github.com/dbhart/lasio/actions

DC

dcslagel:
An initial pull of @dbhart's branch to my fork causes the tests to run but all fail.   :-(
https://github.com/dcslagel/lasio/runs/1099998339?check_suite_focus=true


It looks like the issue is these vcs tool tests.  These were added 8 days ago so test run is probably catching an error in these tests. I'll look into it...
```
tests/test_version.py::test_verify_default_vcs_tool FAILED               [ 86%]
tests/test_version.py::test_explicit_existent_vcs_tool FAILED            [ 87%]
```

dcslagel:
@kinverarity1 @dbhart ,  I've posted pull request https://github.com/kinverarity1/lasio/pull/387 to fix the test_version.py failures related to running GitHub Actions.  If it is approved, it should follow the pull-request. 

I'm still looking at the items that @kinverarity1 had questions about.  I should have some update in a few days. It might be okay to go-ahead with merging these pull-request and have a follow-up pull-request to address any findings from the question items.  

DC

kinverarity1:
@dcslagel OK that would be good. Thank you.

And thank you to @dbhart for investigating and initiating this.

.github/workflows/ci-tests.yml
dcslagel: `--cov-fail-under=80`  will report coverage failure if the test coverage is below 80%.   My current view, is this would be an undesirable constrant on Lasio at this time but go to remember if Lasio continuous integration activity picks up significantly.

dcslagel: `--cov-branch`  is interesting.  It checks coverage on branching code (if/else, try/catch...) 

Reference: https://coverage.readthedocs.io/en/latest/branch.html

However, after tinkering with it,   I found that it is useful for seeing which files have a lot of branching (and thus should be given test attention, but it didn't provide enough information to go right to the branch locations.  So I didn't find it useful for running on every push. Thus I recommend leaving it out of Continuous Integration runs but it is good to run periodically and make test coverage imrprovements based on the files with high numbers of branching logic.

dcslagel:
@kinverarity1 , @dbhart ,

I researched @kinverarity1's review comments and responded to the ones regarding coverage flags.  Also see pull-request #400 for one possible improvement over pip install --editable ".[all]".

Thanks,
DC

.github/workflows/ci-tests.yml
kinverarity1: Thanks

PR #378, opened 2020-07-31
Issue-377: Add writing Nonetype as empty string

dcslagel:
This pull-request is related to issue 377 "Lasio is writing the python None object to the string 'None' when writing HeaderItems to a file"

It implements the suggested solution except with a function instead of a lambda.  The reason is that the function is self-documenting.  If a lambda is preferred let me know and I can make the change.

This pull-request is in draft mode until new testing is added and test failures are fixed:
`FAILED tests/test_write.py::test_write_single_step`

### DIFF
--- a/lasio/writer.py
+++ b/lasio/writer.py
@@ -248,12 +248,23 @@ def write(
 
 
 def standardize_value(value, unit=None):
-    # If an internal representation of a metadata mnemonic has a unit
-    # indicator and a zero value, empty string, or NoneType, then on
-    # lasio.write(...) the value should write a 0 for the value field
-    # instead of "None" or "".
+    """Ensure that 0 is written instead of 'None' for numeric header lines.
+    
+    Args:
+        value (anything): object to be written into the value field
+            of the LAS header line.
+        unit (str): unit for header line.
+        
+    Returns: either 0 (integer) or a string.
+    
+    If an internal representation of a metadata mnemonic has a unit
+    indicator and a value of zero, an empty string, or None, then on
+    lasio.write(...) the value written should be 0 for the value field
+    instead of "None" or "".
+    
+    """
     # value != 0 prevents overwriting a 0.0 value with 0.
-    if unit and not value and value != 0:
+    if (unit) and (not value) and (value != 0):
         value = 0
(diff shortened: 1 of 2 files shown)


dcslagel:
Yeah, the toggle to draft if non-intuitive, its just under 'Reviewers' next to 'Still in progress?'

dcslagel:
Your comment above brings up some questions.  What do you think of these proposals? 

1. On read(), if a file metadata mnemonic has a unit indicator and a zero value it should be represented internally as a zero value instead of NoneType.  

Currently (on master):
reading a las file with  "STEP    .M                 0.0000                :STEP"
sets las.well.step.value to a NoneType
```python
>>> las = lasio.read('tests/examples/single_step_20.las')
>>> las.well['STEP']
HeaderItem(mnemonic="STEP", unit="M", value="None", descr="STEP")
>>> type(las.well.step.value)
<class 'NoneType'>
```

2. If an internal representation of a metadata mnemonic has a unit indicator and a zero value, then on lasio.write(...) the value should write a 0 for the value field instead of "None" or "".

Currently (on master):
If las.well.step.value is set to zero `las.well.step.value = 0`:
This is written out with las.write("newfile.las", version=2.0) as
"STEP.M                    None : STEP"


3. If an internal representation of a metadata mnemonic has a unit indicator and an empty value, the unit indicator infers a numerical value and lasio.write(...) should write a zero `0` as the value field.  

If you agree with these, 2, and/or 3 can be implemented in this pull-request and 1 would be a separate pull request.

kinverarity1:
@dcslagel I agree with all of those ideas - sounds like a good solution!

dcslagel:
Coverage:

```
lasio/examples.py             42     16    62%
lasio/las.py                 388     44    89%
lasio/las_items.py           190     30    84%
lasio/las_version.py          34     22    35%
lasio/reader.py              365     40    89%
lasio/writer.py              168      9    95%
TOTAL                       1325    217    84%
```

Thank you,
DC

kinverarity1 (approved):
Sorry about taking so long to get to this. Excellent, thank you. I will merge now.

Issue #380, opened 2020-09-02
Newest release of 0.26 requires git be installed


choogeGL:
I distributed a program utilizing the newest lasio release (0.26) to users using pyinstaller and they were not able to make the code run.  
It looks like in git is needed - see las_version.py file at Line 71.

In the interim, I've commented this code out and forced the lasio version to be read as 0.26.

Thanks

chooge at geologic.com

dcslagel:
@choogeGL,

Okay, that's a bug.  I'll look into it.

DC

kinverarity1:
@choogeGL Thank you for raising this and apologies for the disruption! 

@dcslagel Thank for having a look. I haven't been able to reproduce the bug yet, but would prefer to get something that works released pending a "proper" fix. So I've opened #381, but am reluctant to merge it since I haven't yet been able to reproduce the bug.

dcslagel:
For the proposed solution in #382,  I was able to reproduce the bug by removing  Git's path from  the `$PATH` variable so that Lasio wouldn't find it. See if it is a good solution.

kinverarity1:
Thank you @dcslagel 👍 

@choogeGL Sorry about the disruption. Version 0.27 on PyPI should resolve the issue with your build. Please re-open the issue here if you encounter the same problem.

PR #381, opened 2020-09-03
Issue #380 v0.26 requires git to be installed - temporary fix

kinverarity1:
This is a temporary fix for the bug #380. I have not been able to reproduce the bug, so this is flying blind a little.


kinverarity1:
See test wheel here for this branch as of de18424: https://github.com/kinverarity1/lasio/releases/tag/0.27.dev3%2Bgde18424

kinverarity1:
I have merged #382 instead - this PR will not be used

PR #385, opened 2020-09-08
Las3.0 : Add registering las30 sections before 1.2/2.0 sections

dcslagel:
#### Changes:
- Register sections containing one of the new 3.0 section indictators:
_Data, _Definition, _Parameter before registering the older Curve and
Parameter sections. This prevents Curves from accidentally parsing
Core_Definitions, Parameter from parsing Perforations_Definition.
- Add test for verifying the above behavior

- This change also adds a Las30_Data stub for future development work in
  parsing Las30 data sections. 

#### Test results:

```
lasio/las.py                 384     55    86%
lasio/las_items.py           190     31    84%
lasio/las_version.py          50     10    80%
lasio/reader.py              445     84    81%
lasio/writer.py              168      9    95%
TOTAL                       1417    255    82%
```

Thank you,
DC


tests/test_read_30.py
```diff
@@ -15,16 +15,22 @@
...
+
+
+def test_read_v30_sample_v20_sections_are_empty():
+    """
+    Verify 'Curves' doesn't read 'Core_*' sections
+    Verify 'Parameter' doesn't read 'Performations_*' sections
+    """
+    las = lasio.read(stegfn("3.0", "sample_3.0.las"))
+    assert las.sections['Curves'] == []
+    assert las.sections['Parameter'] == []
```
kinverarity1, line 34: Maybe this is personal preference but I get worried by ``== []`` - can we change it to something explicit like ``assert not las.sections['Curves']`` or perhaps ``len(las.sections['Curves']) == 0``? Not sure.

lasio/reader.py
```diff
@@ -722,7 +726,14 @@ class SectionParser(object):
     """
 
     def __init__(self, title, version=1.2):
-        if title.upper().startswith("~C"):
+        las3_section_indicators = ['_DATA', '_PARAMETER', '_DEFINITION']
```
kinverarity1, line 729: For now, I think we should only force sections like these to be read explicitly as "LAS 3" sections **if** the parsed version of the file is actually 3.0. If it isn't, there's plenty of chance that for example a genuinely LAS 2 file might have these strings in the section heading for totally different sections.

It's a pedantic point that adds complexity, I know, but I don't want sections that are currently being parsed a certain way to change, at least temporarily. Does that make sense? We're probably not testing the situation I outline in regression tests now, so it would go unnoticed.

lasio/reader.py
```diff
@@ -323,6 +323,10 @@ def determine_section_type(section_title):
         return "Data"
     elif stitle[:2] == "~O":
         return "Header (other)"
+    # las30 transitional code till data parsing is robust for ~A and '_Data'
+    # sections
+    elif stitle.find('_Data') > -1:
+        return "Las30_Data"
```
kinverarity1, line 329: Small point but can we make it `3` instead of `30`?

lasio/las.py
```diff
@@ -210,6 +223,13 @@ def read(
                     logger.debug("Storing reference and returning later...")
                     data_section_indices.append(i)
 
+                # Initial stub for parsing las30 data. This is probably a
+                # transitional section that will merge with 1.2/2.0 data
+                # parsing once fully functional
+                elif section_type == "Las30_Data":
+                    logger.debug("Storing Las30_Data reference and returning later...")
```
kinverarity1, line 230: refer to other comment - small point but can we make it `3` instead of `30`?

lasio/reader.py
```diff
@@ -323,6 +323,10 @@ def determine_section_type(section_title):
         return "Data"
     elif stitle[:2] == "~O":
         return "Header (other)"
+    # las30 transitional code till data parsing is robust for ~A and '_Data'
+    # sections
+    elif stitle.find('_Data') > -1:
```
kinverarity1, line 328: I'm being really picky here but ``> -1`` feels like C - can we use ``elif re.search('_Data', stitle):`` instead? Feels more Pythonic?

kinverarity1 (changes requested):
Thank you. I like the approach, it's a good idea for moving forward. I have some very pedantic changes, sorry 😬

tests/test_read_30.py
dcslagel: Good point!  I think `assert not las.sections['Curves']` would be good and work, but I'll check it out and make a change based on the findings.

lasio/reader.py
dcslagel: Yep, It is a great point.  I think I can add a 3.0 version check to the if statement.

lasio/reader.py
dcslagel: It's a good point, I'll make the change.

lasio/reader.py
dcslagel: Done!, Made the change in 97a72d9.

lasio/las.py
dcslagel: Made the change in 97a72d9.

lasio/reader.py
kinverarity1: Understood, it's a tricky one! Thanks for adjusting.

kinverarity1 (approved):
Thank you

PR #327, opened 2020-05-03
Rearrange reader

kinverarity1:
This PR will finalise rearranging the reader code so that the following process is followed when reading a file:

- [x] 1. The file is iterated over line by line looking for ~ characters and it stores the location of these in the file and the contents of the line (i.e. the section title). 
- [x] 2. The reader returns to each section. For those which are not data sections it seeks to the beginning of each one and parses it by iterating only over the lines which are in that section
- [x] 3. After all that, the parser returns to each data section and parses it, using the information obtaining from parsing the definition/curves section in step 2

PR #393, opened 2020-09-12
Remove reader.py:read_file_contents - unused

kinverarity1:
I think as part of #327 I was probably meant to remove the function `reader.py:read_file_contents`. All tests pass without this function, however coverage is not 100% so I want to be sure before removing the code.

@dcslagel you have probably looked more closely over the related changes than I did; does this seem right to you?


dcslagel:
@kinverarity1,  I have a caution about removing read_file_contents() too quickly also. 

I did the following searches that show Lasio is no-longer calling read_file_contents.
```bash
$ cd lasio
$ grep -rl read_file_contents *
docs/source/lasio.rst
htmlcov/lasio_las_py.html
htmlcov/lasio_debug-reader_py.html
htmlcov/lasio_reader_py.html
lasio/reader.py

$ grep  read_file_contents lasio/reader.py
def read_file_contents(
```

However, In scanning the removing read_file_contents() code, there are some interesting few details we should check whether if they are relevant and if so, exist in the new code.

Could this pull-request be kept around for a while to give us time to read through the code through read through?

Thanks,
DC

kinverarity1:
Yep no hurry 👍

dcslagel:
@kinverarity1, I reviewed reader:read_file_contents() and compared with the current code.  It is okay to proceed with removing reader:read_file_contents().  

Findings:
- https://github.com/kinverarity1/lasio/pull/401  is re-adding a try clause wasn't in the new code but looks like it is still relevant.

dcslagel:
@kinverarity1,

Just checking on this pull-request. I think it is ready to merge, but let me know if it needs changes or is rejected.

Thanks,
DC

kinverarity1:
Thanks @dcslagel

PR #396, opened 2020-09-12
Add check for pushed tag version to version tests

dcslagel:
#### Description:

When getting Lasio's version in GITHUB_WORK flow environment, test for
both a '--no-tag' empty string version and a pushed release tag string
version.

#### Test results (All test pass for all platforms(Linux, Mac, Windows)

regular push:
https://github.com/dcslagel/lasio/actions/runs/251590209

tag push:
https://github.com/dcslagel/lasio/actions/runs/251590303

Thank you,
DC

### DIFF
--- a/tests/test_version.py
+++ b/tests/test_version.py
@@ -8,24 +8,35 @@ version_regex = re.compile(r'^\d+\.\d+')
 
 import lasio.las_version
 
-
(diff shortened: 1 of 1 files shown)


kinverarity1:
Thank you for fixing!

PR #400, opened 2020-09-14
Merge request to gh-ci-changes-1 branch: Change continuous integration to make a lasio wheel

dcslagel:
#### Description: 

    Update github ci pip commands to use wheel
    
    - Add setuptools and wheel to `pip install --update`
    - Remove '--editable' from github ci pip install cmd
    
    This causes lasio to be packaged as a wheel package.  Then the wheel
    file is installed and tested.  Before we were running the tests on the
    editable clone of the lasio repo.

#### Tests

Tab push:
https://github.com/dcslagel/lasio/actions/runs/254618661
```
...
Created wheel for lasio: filename=lasio-0.91-py3-none-any.whl
...
Installing collected packages:...lasio-0.91...
```

Regular branch push:
https://github.com/dcslagel/lasio/actions/runs/254606703
```
...
Created wheel for lasio: filename=lasio-0.1.dev1+g4494276-py3-none-any.whl
...
Installing collected packages:..lasio-0.1.dev1+g4494276...
```

##### Notes:
Because the GitHub-Actions repo is created with --no-tags the Wheel is NOT built with the previous tag. Instead, it uses `0.1` as a default.  However, the commit number `4494276b` is the actual commit of the branch head, which is good.

Thank you,
DC

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -1,5 +1,8 @@
-# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
-# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+# This workflow will install Python dependencies, run tests and lint with a
+# variety of Python versions
+#
+# For more information see:
+# https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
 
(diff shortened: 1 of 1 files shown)


kinverarity1:
Thanks, looks good

Issue #402, opened 2020-09-17
Parse unit starting with dot


eimerej:
I have Las with unit starting with dot:
```TDEP    ..1IN          :0.1-ft Frame Depth {F13.4}```

In this case, the first dot is part of the mnemonic instead of the unit.
This is a special behavior implemented in lasio to support '..' caused by mnemonic abbreviation.
From the function *configure_metadata_patterns(line, section_name)*

```py

if ".." in line and section_name == "Curves":
   # ...
    name_re = name_with_dots_re
```

According to the standard (2.0 and 3.0) mnemonic should not contains dot but unit could.

I see two way of solving this issue:
1. follow the standard and do not use the `name_with_dots_re` regexp
2. if there is some space between the mnemonic and the first dot assume that the dot belong to the unit and not to the mnemonic:
  ```py
  # replace
  if ".." in line and section_name == "Curves":
  #  by something like that
  if re.search('[^ ]\.\.', line) and section_name == "Curves":
  ```

kinverarity1:
Hi @eimerej thank you for raising this issue! Excellent points, and thank you for the PR also!

dcslagel:
Hi @eimerej, @kinverarity1 ,

In your file with `TDEP ..1IN :0.1-ft Frame Depth {F13.4}`  what is the ~WELL INFORMATION section's STRT line?  Is the unit also `.1IN` there?  Could that be used to decide whether the following example has a unit of `1IN` or `.1IN`?
Example: `I. Res..1IN                      : `

Thanks,
DC

dcslagel:
Hi @eimerej, @kinverarity1,

Just thinking through the various scenarios.  While Lasio doesn't currently parse all of these in the following (expected?) ways, 1 - 4 are do-able with just the information in a given metadata line.  Number 5 show situations where how to parse it is ambiguous.

1. `TDEP ..1IN :0.1-ft Frame Depth {F13.4}`
- This could parse clearly  as Name = 'TDEP', Unit = .1
- Pattern [space][dot-delimiter][unit]

2. `TDEP.. 1IN :0.1-ft Frame Depth {F13.4}`
- This could parse clearly  as Name = 'TDEP.', Unit = 1
- Pattern [abbreviation-dot][dot-delimiter][unit]

3. `TDEP. . .1IN :0.1-ft Frame Depth {F13.4}`
- This could parse clearly  as Name = 'TDEP.', Unit = .1
[...]

PR #403, opened 2020-09-17
fix issue when unit starts with dot

eimerej:
Fix for the bug #402

### DIFF
--- a/lasio/reader.py
+++ b/lasio/reader.py
@@ -1052,7 +1052,7 @@ def configure_metadata_patterns(line, section_name):
         if ".." in line and section_name == "Curves":
             name_re = name_with_dots_re
     else:
-        if ".." in line and section_name == "Curves":
+        if re.search(r'[^ ]\.\.', line) and section_name == "Curves":
             double_dot = line.find("..")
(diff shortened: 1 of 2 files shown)


dcslagel:
Draft branch at #413, demonstrates test cases for these examples so you can see the current state of parsing behavior.  The tests make assumptions about the desired behavior that should be checked and 3 of the tests currently fail.

dcslagel:
Hi @eimerej , @kinverarity1 ,

After spending a fair amount of time looking that this, it seems to me that @eimerej 's solution in pull-request https://github.com/kinverarity1/lasio/pull/403  is the right one for now.   It takes care of the cases #1 and #2 that I listed in issue #402 and the cases #3 and #4 are highly unlikely and can be solved if there is ever a real-world case reported.  Case #5, @kinverarity1 , mentioned it is okay to let it resolve with the current logic in Laisio.

So all that said, I encourage merging pull request #403.  Thank you for allowing me to dig in to this problem for an extended period of time.

DC

kinverarity1:
Thank you @eimerej  and @dcslagel - really sorry it has taken us so long here.

PR #424, opened 2021-02-20
Read data section as dataframe

kinverarity1:
This adds a new function to reader.py and makes necessary changes in LASFile so that a data section can be read as a pandas DataFrame. The benefit of this is that now data section columns (== curves) can have different data types, at least when reading unwrapped files. This is happening already such that if items cannot be coerced to floating point numbers, that curve will have a pandas `object` data type i.e. strings.

All tests are passing and apart from the new function in reader.py (read_data_section_into_dataframe) there are no changes to internal function arguments etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,14 +36,14 @@ documentation](https://lasio.readthedocs.io/en/latest/).
 
 ## Quick start
 
-For the minimum working requirements, you'll need numpy installed. Install
-lasio with:
+For the minimum working requirements, you'll need numpy and pandas installed. 
+Install lasio with:
 
 ```bash
 $ pip install lasio
 ```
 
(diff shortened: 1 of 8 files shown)


kinverarity1:
Note to myself: Failing for pandas 0.25.3 - the other tests are using pandas >= 1.

kinverarity1:
Also, I guess for wrapped files now we can attempt to convert each curve to a float after the reader is finished. This will be the way we do it anyway for LAS 3 where the data types are specified.

kinverarity1:
Closing this - based on our work in #452 it's clear that the `data` array should become a record array, with no dependence on pandas.

kinverarity1:
See #461 for the solution to this. No record array, just only ever passing around 1D ndarrays for curves instead.

Issue #404, opened 2020-09-17
lasio changes STEP: adding decimals


geolr:
LAS with a header saying: 
`STEP.m     0.1524 : STEP`
Becomes 
`STEP.m 0.1524000000000001 : STEP`

Only header values were touched in processing, like UWI and adding a prefix to curvenames.
Then las.write to a new file.

Apparently lasio does a re-calculate of depth-step for the first two lines in the data.
I'd assume this has it's origin in floating point arthimetics: Doing the same calculation `<depth-line-2> minus <depth-line-1>` in an interactive Python session give the same result: 0.1524000000000001

Should lasio round the values to 4 decimal places?

kinverarity1:
Hi @geolr thanks for raising this issue! I think there are a few improvements to be made in this area.  Certainly the floating point arithmetic effect there is a bug, I feel like we should represent the depth step at the same precision as present in the depth values themselves.

Your issue has also prompted me to open #405, as I don't think that in your case lasio should be attempting to recalculate the value in the STEP header line at all, as you are only modifying the header.

geolr:
Just a quick note, since I updated: 
lasio-0.28 still has the bug

dcslagel:
Hi @geolr ,

I did this trying to make a test case of this error,  but it passes:

```
def test_step_unchanged_by_write():
    las = read(egfn("2.0/sample_2.0.las"))
    las.well['UWI'] = "123456789"
    las.write('test.las', version=2.0)
    assert las.well['STEP'].value == -0.125
```

As you mentioned above the write function is calculating the STEP from the first two data depths.  In the case of this test it is doing:
```
STEP = 1669.875 - 1670.000   # -0.125
las.well['STEP'].value = STEP
```

- Do you have an example las file that will produce the error?
- How are you adding a prefix to curvenames?  

Thanks,
DC

dcslagel:
analysis notes:

1. running the documentations 'create las file from scratch instructions doesn't show the error:
https://lasio.readthedocs.io/en/latest/building.html
```
import lasio
import numpy as np

testfn = 'testfile.las'
dstart = 10
dstop = 12
dstep = 0.5

las = lasio.las.LASFile()

depths = np.arange(dstart, dstop, dstep)
las.add_curve('DEPTH', depths, unit='m')
las.write(testfn)
```

Outputs:
```
...
~Well ------------------------------------------------------
STRT.m    10.0 : START DEPTH
STOP.m    11.5 : STOP DEPTH
[...]

PR #431, opened 2021-03-18
Adjust precision when calculating STEP on write

dcslagel:
#### Description:

This change addresses "lasio changes STEP: adding decimals" #404

- Use the fractional precision of the 2nd depth value as the precision
  for calculating the STEP value on las.write()
- Add test cases

#### Test Results

All tests pass with a small coverage improvement in las_items.py
```
lasio/las.py                 409     59    86%
lasio/writer.py              177     10    94%
TOTAL                       1411    203    86%
```

--

Note:
There are other ways to decide the fractional precision to use, rounding, floor, ceiling, etc.  Also we could look at more than one depth value to and take the longest, (or average)..  If a another solution will be better let me know and we can make the improvements.

--

Thank you,
DC

### DIFF
--- a/lasio/writer.py
+++ b/lasio/writer.py
@@ -108,11 +108,26 @@ def write(
     if STOP is None:
         STOP = las.index[-1]
     if STEP is None:
+        # prevents an error being thrown in the case of only a single sample being written
         if STOP != STRT:
-            # prevents an error being thrown in the case of only a single sample being written
+            # Try to keep the level of precision that the original data had.
             # Example: 0.51 -> "51" -> len("51") -> 2
             # So we'll make STEP with a fractional len of '2' after the decimal
-            fraction_len = len(str(las.index[1]).split(".")[1])
+            try:
+                fraction_len_0 = len(str(las.index[0]).split(".")[1])
+            except:
+                fraction_len_0 = 0
+
+            try:
+                fraction_len_1 = len(str(las.index[1]).split(".")[1])
+            except:
+                fraction_len_1 = 0
+
+            if fraction_len_0 >= fraction_len_1:
+                fractional_len = fraction_len_0
+            else:
+                fraction_len = fraction_len_1
+
             # Faster than np.gradient
             raw_step = las.index[1] - las.index[0]
             STEP = "{:.{flen}f}".format(raw_step, flen=fraction_len)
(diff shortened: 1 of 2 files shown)


lasio/writer.py
```diff
@@ -110,7 +110,12 @@ def write(
     if STEP is None:
         if STOP != STRT:
             # prevents an error being thrown in the case of only a single sample being written
-            STEP = las.index[1] - las.index[0]  # Faster than np.gradient
+            # Example: 0.51 -> "51" -> len("51") -> 2
```
dcslagel, line 113: Another option is to use the writer.write() parameter: `fmt="%.5f",` instead of calculating the length.

lasio/writer.py
```diff
@@ -110,7 +110,12 @@ def write(
...
+            # So we'll make STEP with a fractional len of '2' after the decimal
+            fraction_len = len(str(las.index[1]).split(".")[1])
```
kinverarity1, line 115: does this work when the index curve is in integers?

kinverarity1 (changes requested):
Thanks - see review comment.

lasio/writer.py
dcslagel: Good catch, I'll fix that for the case of integers and/or decimal mark is missing and post another commit.

dcslagel: I've Added additional checks for fractional section of number.  If it does not exist, then fraction_len is defaulted to 0.

Also I added a check of the initial las.index[0] to make sure we get the largest fractional_len of the two numbers.

This approach is an attempt to retain the fractional len of the original data.  If this is an important goal, then the additional complexity may be worth it.  

I'll make a 2nd pull request that standardizes output on `fmt="%.5f"` for comparison.  To me, the `fmt="%.5f"` feels more reliable and maintainable. It has a sensible default and allows the user to pass in a different value for `fmt`.

lasio/writer.py
dcslagel: I've make a 2nd pull request #432 that standardizes output on `fmt="%.5f"` for comparison.   It currently fails on some older tests because it changes the fractional precision of the STEP value.

dcslagel: Tests in #432 have been updated to pass.

lasio/writer.py
dcslagel: #432 the failing test have been updated to match `.5f` precision when writing the STEP value.  The tests pass now.

dcslagel:
If the solution in pull-request #432 is accepted this solution can be closed.

kinverarity1:
Closed as #432 was merged instead.

PR #448, opened 2021-04-16
quick and dirty

Boorhin:
some very quick idea on how to use the numpy reader


kinverarity1:
Thanks Julien - I'll close this since we are settling on the path in #452 👍

PR #450, opened 2021-04-17
Add data section reader which uses pandas.read_csv

kinverarity1:
This PR replaces `reader.py:read_data_section_iterative` with `pd.read_csv`, essentially.

improves speed for the test file (~29,900 rows x 8 columns) from 1.65 sec to 0.73 sec.

```
--------------------------------------------------------------------------------------------------- benchmark: 2 tests ---------------------------------------------------------------------------------------------------
Name (time in ms)                                  Min                   Max                  Mean              StdDev                Median                 IQR            Outliers     OPS            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_read_v12_sample_big (NOW)                728.1707 (1.0)        986.4181 (1.0)        805.3038 (1.0)      104.4843 (4.71)       769.9727 (1.0)
    102.9696 (3.18)          1;0  1.2418 (1.0)           5           1
test_read_v12_sample_big (0001_d2e0ca0)     1,654.0345 (2.27)     1,710.7092 (1.73)     1,679.1493 (2.09)      22.1773 (1.0)      1,672.4063 (2.17)
     32.3871 (1.0)           2;0  0.5955 (0.48)          5           1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

However 21 tests are failing, mostly due to the non-application of substitutions and so on. Will need to work out a way to read these files using the normal reader...

```
FAILED tests/test_null_policy.py::test_null_policy_ERR_custom - TypeError: ufunc 'isnan' not supported for the input types, and the inputs could ...
FAILED tests/test_null_policy.py::test_null_policy_text_all_subs_ERR - TypeError: ufunc 'isnan' not supported for the input types, and the inputs...
FAILED tests/test_null_policy.py::test_null_policy_text_all_subs_null - TypeError: ufunc 'isnan' not supported for the input types, and the input...
FAILED tests/test_null_policy.py::test_null_policy_text_dashes_1 - TypeError: ufunc 'isnan' not supported for the input types, and the inputs cou...
FAILED tests/test_null_policy.py::test_null_policy_text_dashes_2 - TypeError: ufunc 'isnan' not supported for the input types, and the inputs cou...
FAILED tests/test_null_policy.py::test_null_policy_text_dashes_3 - AssertionError: assert '-2550.000' == -2550
FAILED tests/test_null_policy.py::test_null_policy_runon_replaced_1 - TypeError: ufunc 'isnan' not supported for the input types, and the inputs ...
FAILED tests/test_null_policy.py::test_null_policy_runon_replaced_2 - AssertionError: assert False
FAILED tests/test_null_policy.py::test_null_policy_runon_ok_1 - AssertionError: assert '7.330-19508.961' == 7.33
FAILED tests/test_null_policy.py::test_null_policy_runon_ok_2 - assert 10.424 == -19508.961
FAILED tests/test_null_policy.py::test_null_policy_small_non_zero_neg_nums - AssertionError: assert False
FAILED tests/test_read.py::test_read_v12_sample_wrapped - lasio.exceptions.LASDataError: Traceback (most recent call last):
FAILED tests/test_read.py::test_read_v2_sample_wrapped - lasio.exceptions.LASDataError: Traceback (most recent call last):
FAILED tests/test_read.py::test_comma_decimal_mark_data - AssertionError: assert '123,420' == 123.42
FAILED tests/test_read.py::test_missing_a_section - lasio.exceptions.LASDataError: Traceback (most recent call last):
FAILED tests/test_read.py::test_blank_line_in_header - lasio.exceptions.LASDataError: Traceback (most recent call last):
FAILED tests/test_read.py::test_skip_comments_in_data_section - AttributeError: 'bool' object has no attribute 'all'
FAILED tests/test_read.py::test_quoted_substrings_in_data_section - lasio.exceptions.LASDataError: Traceback (most recent call last):
FAILED tests/test_wrapped.py::test_wrapped - lasio.exceptions.LASDataError: Traceback (most recent call last):
FAILED tests/test_wrapped.py::test_write_wrapped - lasio.exceptions.LASDataError: Traceback (most recent call last):
FAILED tests/test_wrapped.py::test_write_unwrapped - lasio.exceptions.LASDataError: Traceback (most recent call last):
============================================ 21 failed, 214 passed, 2 skipped, 8503 warnings in 26.79s ============================================
```

### DIFF
--- a/lasio/reader.py
+++ b/lasio/reader.py
@@ -7,7 +7,6 @@ import traceback
 import urllib.request
 
 import numpy as np
-import pandas as pd
 
 from . import defaults
 
@@ -394,7 +393,7 @@ def inspect_data_section(file_obj, line_nos, regexp_subs, remove_line_filter="#"
         return item_counts[0]
 
 
-def read_data_section_iterative(
+def read_data_section_iterative_normal_engine(
     file_obj, line_nos, regexp_subs, value_null_subs, remove_line_filter
 ):
     """Read data section into memory.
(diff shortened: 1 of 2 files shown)


kinverarity1:
I have now made the dependency on pandas "soft" and the parser is invoked with the option keyword argument on `LASFile.read` `engine='pandas'`.

kinverarity1:
Whoops - tests failing on python 3.5 because I used some f-strings. Otherwise - tests should be passing. I still need to add some tests to bring coverage back up and add some mentions to docs.

The next step is to allow different data types in each column (e.g. like #424)

lasio/reader.py
```diff
@@ -454,6 +454,55 @@ def items(f, start_line_no, end_line_no):
...
+        remove_startswith (str): reject all lines starting with that value e.g. ``"#"``
+
+
+    Returns:
+        A 1-D numpy ndarray.
+
+    """
+    import pandas as pd
+
+    logger.debug(f"regexp_subs: {regexp_subs}")
```
kinverarity1, line 479: Remove f-string

kinverarity1 (commented):
See comments

Boorhin:
OK sorry but I am really struggling with accessing the draft branch. 
what you use in the panda version 
`    array = pd.read_csv(file_obj, skiprows=0, header=None, nrows=nrows, delim_whitespace=True).values`
Could be simply replaced by:
`array = np.genfromtxt(file_obj)`
There is no need to tell numpy to stop reading as soon as the data will fail, if you have built an artificial object file (file_obj) there is no need to tell to skip anything. I think that it is a bit of a shame to not structure your array with the column name like I suggested in my tiny example. This will allow you to querry from the header dictionary your data in the array. 
genfromtxt will recognise the ASCII delimiters by default (consecutive white spaces).
I think that it is strategically not great to build the file_obj I think I would need to look into this in more details. It is better to get your header data and then read your file from the same access point in memory with the "skip_row and names arguments".

dcslagel:
Since the 9 wrapped tests fail.  A couple of options we could do here are once a LAS file has been identified as wrapped: change the engine from 'pandas' to 'normal'  or unwrap the file and return the unwrapped instance the the pandas engine.  Are either of these a good approach?

kinverarity1:
I think we stick with using the normal reader per discussion in swung slack. Closing to focus our work in #452

Issue #268, opened 2019-02-23
Remove side-effects from write()


kinverarity1:
At the moment there are side-effects to the ``lasio.writer.write()`` function involving modifications to the LASFile objects' STRT, STOP and STEP values.

I should avoid these side-effects entirely - see previous discussion in #35. They would better belong in a ``LASFile.update_start_stop_step()`` method, or something similar.

dcslagel:
If #447 is accepted and merged then there are these next decisions and steps to completing this task:

- Move `las.update_start_stop_step()` out of `writer.write()`
  This means that `writer.write()` will no-longer automatically change the format of STRT, STOP and STEP

- Update the tests that will break.  Should these tests include `las.update_start_stop_step()` or not?
```
FAILED tests/test_wrapped.py::test_write_wrapped
FAILED tests/test_wrapped.py::test_write_unwrapped
FAILED tests/test_write.py::test_write_sect_widths_12
FAILED tests/test_write.py::test_write_sect_widths_20_narrow
FAILED tests/test_write.py::test_write_sect_widths_20_wide
FAILED tests/test_write.py::test_df_curve_addition_on_export
FAILED tests/test_write.py::test_multi_curve_mnemonics_rewrite
FAILED tests/test_write.py::test_multi_curve_missing_mnemonics_rewrite
FAILED tests/test_write.py::test_write_units
FAILED tests/test_write.py::test_rename_and_write_curve_mnemonic
FAILED tests/test_write.py::test_write_single_step
FAILED tests/test_write.py::test_write_12_to_20_ver_in_mem_is_12
FAILED tests/test_write.py::test_write_12_to_20_ver
FAILED tests/test_write.py::test_write_20_to_12_ver_in_mem_is_20
FAILED tests/test_write.py::test_step_unchanged_by_write
FAILED tests/test_write.py::test_step_unchanged_by_write_2
```

- New tests for using `las.update_start_stop_step()`


- Update documentation
  - include `las.update_start_stop_step()` documentation/examples
  - update `writer.write()` documentation

kinverarity1:
Hmm my feeling is that maybe we should track, in LASFile, whether the user has modified the index curve. If they have, then LASFile.write should call update_start_stop_step() - if they haven't, it shouldn't....?

dcslagel:
Okay, thinking how to do track changes to the index curve.  I'll work on that next.

PR #451, opened 2021-04-17
On write, update step vars if curve idx changed

dcslagel:
#### Description:

This should complete Remove side-effects from write() #268.

This is a first iteration at updating the STEP variable as part of the write() process but only in the following cases:
1. an initial curve index was not read from a las file (las.index_initial)
2. or the curve index has changed during processing
3. or if the STOP value doesn't match the final index value

I am not sure if case #3 should trigger updating the step variables.  It looks like many of the test case LAS files have STOP values that don't match the final index value or the final index value - 1 STEP.   Let me know if these needs to be modified.

#### Test Results:
Python 3.9

```
lasio/las.py                 431     60    86%
lasio/writer.py              166     10    94%
TOTAL                       1422    204    86%
```

--

Thank you,
DC

### DIFF
--- a/lasio/las.py
+++ b/lasio/las.py
@@ -65,6 +65,8 @@ class LASFile(object):
         super(LASFile, self).__init__()
         self._text = ""
         self.index_unit = None
+        self.index_name = None
+        self.index_org = None
         default_items = defaults.get_default_items()
         self.sections = {
             "Version": default_items["Version"],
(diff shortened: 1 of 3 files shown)


kinverarity1:
1. - STEP should definitely be auto-calculated if LASFile is created from scratch.

2. I'll look at the failing tests tonight but think that we can safely change them as long as the expected output makes sense...

kinverarity1 (approved):
Looks great, thanks!

Issue #453, opened 2021-04-19
rounding issues when writing las file


eucalypt1:
When saving a lasfile using las.writer(filename, version=2),  the following well items show as shown below:
STRT.M              1240.0 : START DEPTH
STOP.M  1258.1999999999834 : STOP DEPTH
STEP.M 0.09999999999990905 : STEP

No matter the date frame data or what I specify prior to running las.writer, the data shows as above. I try to show them as:
STRT.M              1240.0 : START DEPTH
STOP.M             1258.2  : STOP DEPTH
STEP.M                  0.10 : STEP

Also, when creating a lasfile using:  las = lasio.LASFile() and try to save it later with las.writer, if the other sections (Curves, Params and the actual data values (below ~ASCII) are not filled, the las,writer returns an exception. I use visual studio  (if that plays a role in this issue)

Thanks for all your work.

dcslagel:
Hi @eucalypt1,   

What Lasio version are you using?

Could you include a minimum code example that reproduces the problem you are seeing with las.writer()?

Thank you

eucalypt1:
[...]

PR #455, opened 2021-04-21
Some updates to rounding behavior when writing for issue 453

dcslagel:
#### Description:

This pull-request resolves the issues identified in `rounding issues when writing las file` #453.

1. Updates for rounding behavior when writing:
    
- If we are getting STRT and STOP from the data then format them to a
  standard precision. Current default precision is "%.5f".
- If STRT and STOP are passed in with values, don't format them because
  we assume they are at the user's expected precision.
- Update failing test cases to have the correct STOP value.
- Add a test case for the case where a LAS file does not have a correct
  STOP value.
- Move updating units to a separate LASFile method so that units are
  always synced between las.curves.unit and las.well.<step>.unit.

2. Enable writing empty LAS file:

- Handle exceptions on empty data
- Add test case for writing an empty LAS file: test_write_empty_las()

#### Testing Results:

```
lasio/las.py                 436     59    86%
lasio/writer.py              171      9    95%
TOTAL                       1432    202    86%


- benchmark: 1 tests -------------------------------
Name (time in s)                Min     Max    Mean  
----------------------------------------------------
test_read_v12_sample_big     944.9445  957.4294  950.0452
---------------------------------------------------
```

--

Thank you,
DC

### DIFF
--- a/lasio/las.py
+++ b/lasio/las.py
@@ -373,7 +373,20 @@ class LASFile(object):
             self.index_initial = self.index.copy()
 
     def update_start_stop_step(self, STRT=None, STOP=None, STEP=None, fmt="%.5f"):
-        """Configure or Change STRT, STOP, and STEP values"""
+        """Configure or change STRT, STOP, and STEP values on the LASFile object.
+
+        Keyword Arguments:
+            STRT, STOP, STEP (str, int, float): value to set on the relevant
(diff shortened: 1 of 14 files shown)


kinverarity1 (approved):

kinverarity1:
Looks great thank you.

PR #456, opened 2021-04-23
Remove unnecessary file_obj.tell() calls

kinverarity1:
Thanks @donald-keighley for noting this.


kinverarity1:
Scratch that - not as simple a change as I anticipated.

PR #452, opened 2021-04-18
Add a numpy engine for reading using numpy.genfromtxt()

dcslagel:
#### Description:

This draft branch is to implement using `numpy.genfromtxt()` as an alternate data reader.  It is work for issue #446 `Use an accelerated numpy or pandas reader`.
 
- It builds on the work in `Add data section reader which uses pandas.read_csv #450` 
- It defaults to engine="numpy" for testing purposes. 
- Adds and separate function : `read_data_section_iterative_numpy_engine()`
- When a file indicated that it is wrapped, then it will be read with engine="normal"  instead.

#### Test Results:

Highlights:
- The speed test is significantly faster than the master branch
- Overall test coverage  is declining, only down 1% from 86% to 85%.
- There are currently **21** test failures.  These generally divide into AssertionErrors and lasio.expections.LASDataErrors.

```
---------- coverage: platform darwin, python 3.9.4-final-0 -----------
Name                       Stmts   Miss  Cover
----------------------------------------------
lasio/__init__.py             13      2    85%
lasio/convert_version.py      20     20     0%
lasio/defaults.py             11      0   100%
lasio/examples.py             42     10    76%
lasio/excel.py                88     34    61%
lasio/exceptions.py            6      0   100%
lasio/las.py                 451     65    86%
lasio/las_items.py           199     29    85%
lasio/las_version.py          50     14    72%
lasio/reader.py              446     45    90%
lasio/writer.py              171      9    95%
----------------------------------------------
TOTAL                       1497    228    85%

```

##### Benchmark comparison with Master branch: numpy-genfromtxt-explore is significantly faster
```
------------------------------------------------------------------------------------------------- benchmark: 2 tests -------------------------------------------------------------------------------------------------
Name (time in ms)                                  Min                   Max                  Mean            StdDev                Median               IQR            Outliers     OPS            Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_read_v12_sample_big (NOW)                366.4775 (1.0)        373.0084 (1.0)        370.3707 (1.0)      2.9769 (1.68)       371.9692 (1.0)      5.1055 (1.62)          1;0  2.7000 (1.0)           5           1
test_read_v12_sample_big (0001_1c1220f)     1,087.4714 (2.97)     1,091.4720 (2.93)     1,089.7887 (2.94)     1.7711 (1.0)      1,090.2695 (2.93)     3.1479 (1.0)           1;0  0.9176 (0.34)          5           1
[...]

Issue #416, opened 2021-01-06
Time-indexed LAS file


LuisNavarro93:
Is there a way to import .las file with Date-Indexed data??
![image](https://user-images.githubusercontent.com/62827729/103789972-3c805000-5006-11eb-9e8f-b322aa58f1b5.png)
The LAS file structure goes as follows.
![image](https://user-images.githubusercontent.com/62827729/103790225-8b2dea00-5006-11eb-936b-efc1a5a315b1.png)
![image](https://user-images.githubusercontent.com/62827729/103790273-9a149c80-5006-11eb-9a0e-015f3bc52b65.png)

kinverarity1:
Thanks for raising this issue. It's not possible at the moment, but will be part of LAS 3.0 support when we implement that.

LuisNavarro93:
Thank u... Looking forward to the release

kinverarity1:
Could you please try this again with lasio v0.30?

It should now work: https://lasio.readthedocs.io/en/v0.30/data-section.html#handling-text-dates-timestamps-or-any-non-numeric-characters

Issue #227, opened 2018-03-13
NULL value replacing valid value in DEPT


kwinkunks:
Since DEPT is treated as a curve, I guess it gets clobbered by the NULL value:

![image](https://user-images.githubusercontent.com/1692372/37361438-a3bc78e0-26d1-11e8-908e-2350f31acebb.png)

I thought the default NULL value is -999.25, so maybe there are two problems here... Not sure.

I'm on version 0.19.

kinverarity1:
Thanks for the report. 

Yes, lasio is considering 999.25 as a NULL in your case because of the default value for null_policy (``lasio.read(..., null_policy='common')`` - see the [documentation](http://lasio.readthedocs.io/en/latest/data-section.html#handling-invalid-data-indicators-automatically)). If you use ``null_policy='strict'`` that will fix your problem.

From a larger perspective this is tricky to fix because the replacements are done before wrapping is applied - see https://github.com/kinverarity1/lasio/blob/master/lasio/reader.py#L328. If I choose to ignore null substitutions for the first column, that's easy enough, but will only work for unwrapped files. I'd certainly welcome a PR that ignores all substitutions for the true first column in both wrapped and unwrapped files!

VelizarVESSELINOV:
Suggestion to replace NULL after the reading of the file (wrapped or unwarapped) and skip the index column.

kinverarity1:
Yes that would be ideal... PRs welcome ;-)

kinverarity1:
FYI this is now fixed in `master`:

![image](https://user-images.githubusercontent.com/4525931/116016441-c3237080-a67b-11eb-899a-2f56c63531be.png)

PR #461, opened 2021-04-25
Allow different data types per curve in data section reader

### LINKED ISSUE #439: Data section containing non-numeric chars is parsed entirely as `str`
**Describe the bug**
If the data section contains any non-numeric characters, all of it is parsed as `str`, instead of data types being per-column. This is the fundamental blocker to supporting many LAS 3 files in a useful way and has been touched on before e.g. https://github.com/kinverarity1/lasio/issues/5#issuecomment-685567944

**To Reproduce**
- See attached file: 
[test_utah_well.las.txt](https://github.com/kinverarity1/lasio/files/6289410/test_utah_well.las.txt)
- Sourced from https://gdr.openei.org/submissions/1076

**Expected behavior**
Each curve in `las.curves` should have an appropriate data type. For the attached example, the curves `Colour` and `Type` should be type `str` and all the others should have type `float`, effectively.

**Software versions (please complete the following information):**
 - Python version: N/A
 - Python distribution N/A
 - lasio version (`import lasio; print(lasio.__version__)`) `master`

kinverarity1:
This commit changes the data section reader function (`reader.py:read_data_section_iterative`) to return a generator which yields the data ndarray (1D) for each curve in turn. The old behaviour was for the function to return a 2D ndarray.

It also:

- reshaping of the data array to 2D now occurs solely inside  the data section reader function, which has a new argument   "n_columns" to allow this to happen
- adds dtypes kwarg to `las.py:LASFile.read`
- add `reader.py:identify_dtypes_from_data` function to implement  dtypes="auto" (default value) by using the first row of
  data to automatically identify column data types
- changes to `las.py:LASFile.read` to cater for the above

Consequences:
- This PR will fundamentally resolve #439 Data section containing non-numeric chars is parsed entirely as str
- Provides a basis for fixing e.g. #416 Time-indexed LAS file
- Possibly partially resolve (and provide a way to fix) #227 NULL value replacing valid value in DEPT 
- @dcslagel Eventually we will need to merge this into PR #452 (numpy engine) or rebase - I'm happy to help out with this.

Details:
- Coverage increased from 85% to 86%
- Benchmark test from 1.4 s to 1.0 s ?

```
---------- coverage: platform linux, python 3.6.13-final-0 -----------
lasio/las.py                 419     60    86%
lasio/las_items.py           190     29    85%
lasio/las_version.py          50     15    70%
lasio/reader.py              435     27    94%
lasio/writer.py              171      9    95%
TOTAL                       1445    206    86%
Coverage XML written to file coverage.xml


----------------------------------------------- benchmark: 1 tests ----------------------------------------------
Name (time in s)                Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------
test_read_v12_sample_big     1.0652  1.0758  1.0700  0.0041  1.0694  0.0062       2;0  0.9346       5           1
-----------------------------------------------------------------------------------------------------------------
```


lasio/las.py
```diff
@@ -279,68 +300,31 @@ def read(
...
-                            arr[arr == provisional_null] = np.nan
-                        logger.debug("~A after NULL replacement data {}".format(arr))
-
-                        # Provisionally, assume that the number of columns represented
-                        # by the data section's array is equal to the number of columns
-                        # defined in the Curves/Definition section.
 
-                        n_columns_in_arr = len(self.curves)
+                    # Assign data to curves.
+                    n = 0
```
dcslagel, line 305: Could the name of the variable 'n' be changed to something more descriptive?  Something like curve_array_idx.  This helps others understand the code and helps with searching for the variable.

dcslagel (changes requested):

dcslagel (commented):
This change looks good.  I had one small change request.

lasio/las.py
kinverarity1: Thanks, good point, I've renamed.

Issue #375, opened 2020-07-25
Provide curve names as header row in ~A line


kinverarity1:
Provide an option in `LASFile.write` to put curve names in the ~A row.

This is related to #30 and #90 which address reading this kind of data in.

PR #466, opened 2021-05-18
Add tests for data-section-header column names

dcslagel:
#### Description:

This pull-request adds test for the data-header-line branch change.   This duplicates the examples in `data-section-header.ipynb` at https://gist.github.com/kinverarity1/aaf3459c3fa3eeafe127c9011c15c835.

```
tests/test_write.py::test_write_data_section_header_default
tests/test_write.py::test_write_data_section_header_with_curve_names
tests/test_write.py::test_write_data_section_header_renamed_with_curve_names 
```

#### The edge cased described in #375 are not covered in these tests:
- where mnemonics are missing from the ~C section, there should be the appropriate lasio-generated UNKNOWN mnemonic in the ~A header row only (not in the ~C section)
-  where duplicate mnemonics exist, the ~A header row should contain these as duplicates, NOT the lasio-generated GAMM[1], GAMM[2] etc with suffixes
-  where mnemonics are longer than the specified write fmt, the width of the column in the data section will need to be increased to cater for the increased length of the mnemonic
-  when the data section is wrapped, specifying mnemonics_header_row=True should revert to False, and a logger.warning call made to inform the user/application that this doesn't work.

### DIFF
--- a/tests/test_write.py
+++ b/tests/test_write.py
@@ -780,3 +780,162 @@ API .          : API NUMBER
 ~ASCII -----------------------------------------------------
 """
     )
+
+
+def test_write_data_section_header_default():
+    # Verify default behavior
+    s = StringIO()
+    las = read(egfn("2.0/sample_2.0.las"))
+    las.write(s)
+    s.seek(0)
+    assert (
+        s.read()
+        == """~Version ---------------------------------------------------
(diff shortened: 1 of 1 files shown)


kinverarity1:
Thank you!

Issue #486, opened 2021-09-10
Adding a citation file, and maybe a DOI


ThomasMGeo:
**Is your feature request related to a problem? Please describe.**
No current problem with lasio, I would like to cite it in an upcoming manuscript! 

**Describe the solution you'd like**
https://guides.github.com/activities/citable-code/ is a great start!

**Additional context**
Thanks!

dcslagel:
@ThomasMGeo ,

CITATION.cff is now available in the GitHub repo.  It will be available in a release on the next release.  Could you check and see if that satisfies this issue?

Thank you,
DC

ThomasMGeo:
I think this looks great! Thanks! :)

PR #487, opened 2021-09-10
Add initial CITATION.cff

dcslagel:
#### Description:

This pull-request is for Issue #486 "Adding a citation file, and maybe a DOI".  
This change adds an initial CITATION.cff file.  However, I didn't find a DOI so the file does not include a DOI.  
If a DOI is created in the future it can be added to the CITATION.cff at that time.

--

Thank you,
DC

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -6,6 +6,8 @@ authors:
   - family-names: Inverarity
     given-names: Kent
     orcid: https://orcid.org/0000-0002-5902-2007
+  - name: "Lasio contributors"
+    website: "https://lasio.readthedocs.io/en/latest/contributing.html"
 repository-code: "https://github.com/kinverarity1/lasio"
 version: 0.29
 url: https://github.com/kinverarity1/lasio/releases/tag/v0.29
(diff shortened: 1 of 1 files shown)


kinverarity1 (commented):
Looks good, but I think the authors should include all contributors, not just me. Is there a standard practice for software citations?

dcslagel:
The following article is what I could find:
https://peerj.com/articles/cs-86/

I read it in the next couple of days.... and have an update.

Also, I did look at several CITATION.cff examples and the tended to have a small a number of authors much like a research-article.

dcslagel:
Update:
I updated the pull request with option 4 in the discussion above since I don't know the family-name and given-name of some of the contributor since they contribute using their GitHub id name and this solution seems like a sensible balance between to include contributions without needing to update this file each time there is a new contributor.

I used the Lasio documentation link to the contributor rather than the GitHub url to contributors, https://github.com/kinverarity1/lasio/graphs/contributors, but both  seem like reasonable options.  Let me know if the GitHub link is preferred and I'll make the change. 

--

Thank you,
DC

kinverarity1 (approved):
Great! Thank you!

Issue #392, opened 2020-09-12
ignore_comments not mentioned in documentation


kinverarity1:
The keyword argument `ignore_comments` on `las.py:LASFile.read` should be mentioned in the documentation, probably here: https://lasio.readthedocs.io/en/latest/header-section.html#handling-errors.

https://github.com/kinverarity1/lasio/blob/90a4dacabad1ee564eb8be39cb87a765f38ac8bf/lasio/las.py#L83-L125

PR #495, opened 2021-11-01
Add ignore_comments to documentation

dcslagel:
#### Description:

Add ignore_comments to documentation.  This should resolve issue: "ignore_comments not mentioned in documentation" #392.

Also change data parsing functions to use 'ignore_data_comments' rather than 'ignore_comments'. This clarifies the difference between the two flags since 'ignore_data_comments' is used in parsing the data section and 'ignore_comments' is used in parsing header sections.

--

Thank you,
DC


kinverarity1 (approved):
Looks great. Thank you!

PR #500, opened 2021-11-08
Fix missing docs example for "Mnemonics which contain a period"

kinverarity1:
image

### DIFF
--- a/docs/source/header-section.rst
+++ b/docs/source/header-section.rst
@@ -339,7 +339,7 @@ Mnemonics which contain a period
 As with other LAS file parsers, lasio does not parse mnemonics which contain
 a period - instead, anything after the period will be parsed as the unit:
 
-.. code-block::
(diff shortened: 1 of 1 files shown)


dcslagel:
@kinverarity1,

This doesn't look like it is updated/corrected at https://lasio.readthedocs.io/en/latest/header-section.html?highlight=period#mnemonics-which-contain-a-period

However it is updated (the example displays) if I build the docs locally.

DC

dcslagel:
@kinverarity1,

One additional thing, which may just be a problem with my local setup so just checking.  When I run rstcheck, or rest2pseudoxml on the rst files they display a Warning/Error for the `.. code-block:: none`  entries:
`Cannot analyze code. No Pygments lexer found for "none"`.

Do you get these messages also?  If you don't see them, then I have some misconfiguration in my document-building environment.

DC

PR #517, opened 2022-06-15
Remove Python 3.5 and add 3.9 and 3.10 to GitHub Actions workflow

kinverarity1:


dcslagel:
It looks like quotes are needed around 3.10:
per: https://github.com/actions/setup-python/issues/160
```
3.10 is parsed as a number and it is trimmed to 3.1 that looks like expected behavior for numbers.
You should specify input with quotes to treat it as string.

version: '3.10'
```

kinverarity1:
thanks for your help with this @dcslagel

PR #518, opened 2022-06-16
Pull-Request to ci-changes  branch for python "3.10" and black install

dcslagel:
#### Description: 

Add black to setup's TEST_REQS and fix ci-change for python "3.10"

  - This change includes the ci-change to remove python 3.5 from GitHub Action testing
    and add 3.9 and 3.10.
  - Fix: Add quotes around 3.10 in ci-test.yml so that 3.10 is not parsed as 3.1.
  - Now black will be installed when running 'pip install ".[test]", which
    installs packages needed for developing Lasio.
  - The install black step is removed from the contributing doc.
  - Remove python 3.2 - 2.5 from setup.py classifiers
  - Add python 3.9 and 3.10 to setup.py classifiers

--

Thank you,
DC

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9, 3.10]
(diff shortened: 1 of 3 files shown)


kinverarity1:
Thank you for fixing this :-)

Issue #394, opened 2020-09-12
Update API documentation to reflect current contents


kinverarity1:
The API documentation is probably not including some new or renamed functions, and perhaps includes some unused ones. Needs updating.

https://lasio.readthedocs.io/en/latest/lasio.html

https://github.com/kinverarity1/lasio/blob/master/docs/source/lasio.rst

dcslagel:
@kinverarity1, Is there anything I've missed updating in the document pull-request earlier in the year that should be updated in the API documentation?

Thanks!,
DC

dcslagel:
I'm not sure if this is done now, but I've removed myself as the assignee for the time being.

PR #523, opened 2022-07-01
Docs: Add a few newer API functions

dcslagel:
#### Description:

This pull-request addresses part of issue #394 "Update API documentation to reflect current contents" by adding a few of the newer functions and methods to the API documentation.

This was done manually by scanning the files.  Some of the newer functions/methods seemed to be helper-functions rather than API related. So I left them out of this commit.  Let me know if they all should be included and update this change to include them.

Also it does look like there are automated ways of generating the API docs.  Did you use one of those techniques on earlier updates to the API?  If so, then maybe this pull-request should be redone using the same method.

--

Thank you,
DC


kinverarity1 (approved):
Really great - thank you for updating!!

PR #524, opened 2022-07-07
Update API Docs: Another approach: standardized API doc generation

dcslagel:
#### Description:

This pull-request is a companion option to #523 "Docs: Add a few newer API functions".  This is a different approach and may not be desirable or might suggest and even better approach.  The main changes are:

- Experiment with a standardized, more generic lasio.rst.  This option replaces the topic organized lasio.rst with a generalized one that calls the same sphinx config on each of the lasio module files and documents their doc strings.  Both approaches have benefits.  
     - The original lasio.rst is organized by topic: reading las files, displaying data, writing las files, etc.  This is helpful to the user.  However, the lasio.rst needs to be manually updated to add new API calls.
     - This example lasio.rst, doesn't require manual updating. Each time the docs are generated the latest API information is used.  However, the module layout isn't as human user friendly.

- Adopt some of the organization style of welly in index.rst. Mainly add Section captions to the document layout and some project related links

--

Let me know if this change is preferred to #523,  or if #523 is preferred or if there is a hybrid approach that enables the benefits of both of #523 and this pull-request.  I think additional changes to either of these options is  needed to get the right scope of detail and organization for the API.

Thank you,
DC

--

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -34,7 +34,7 @@ from pkg_resources import get_distribution
 release = get_distribution("lasio").version
 version = ".".join(release.split(".")[:2])
 
-language = None
+language = "en"
 
 exclude_patterns = []
 
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -46,8 +46,10 @@ with Python 2.7 support is version 0.26.
 .. |Research software impact| image:: http://depsy.org/api/package/pypi/lasio/badge.svg
    :target: http://depsy.org/package/python/lasio
 
+
 .. toctree::
-   :maxdepth: 5
+   :maxdepth: 2
+   :caption: User guide
 
    installation
    basic-example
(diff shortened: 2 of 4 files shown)


dcslagel:
I'm closing this pull request.  It can be re-opened if there is interest.  For now I think that more https://github.com/kinverarity1/lasio/pull/523 "Docs: Add a few newer API functions"  pull request builds more naturally on the existing Lasio documentation and should be used.

kinverarity1:
Thanks for this - sorry was just a bit busy this past week. I think it's really good, aligning with welly makes sense. Plus, not needing manual updating is brilliant. Any reason not to re-open and merge this, do you think?

dcslagel:
I've added the Welly style section headers to an upcoming branch and will have a pull request for it in a couple of days.

I'm still looking into the different ways to format and organize an automated API updating configuration.  The current manual arrangement in docs/source/lasio.rst into different task areas (reading, viewing, writing, and creating) makes sense for human readers.  My initial try at automating the API parsing in this pull-request creates an output that doesn't seem as user friendly.   So I'm still researching the options and what other projects are doing.   If I don't find anything better, then we can revive this pull-request.  I'm working on this on-an-off during the next couple of weeks and during that time-range I'll have an update on whether there is a better option or not.

PR #555, opened 2023-02-15
Fix problem when writing with changed data (different number of depths)

bgschaid:
Comparing two numpy arrays has different results
- if they are of the same size: an array of bool (this is what the code expects)
- if they are of different size: the bool value False 
The numpy-function array_equal unifies this behaviour: if they are of the same size it only returns True if all values are the same

Without the fix the test fails at line 25 (las.write)

Put the test in a separate file because my editor kept breaking some other tests (stripped the trailing spaces from some of the multi-line strings)


kinverarity1 (approved):
Thank you @bgschaid and @dcslagel!

PR #571, opened 2023-04-07
Add extras to test requirements

kinverarity1:
Note that tests don't run with `pip install -e .[test]` because some tests require openpyxl. Therefore, test requirements should be both these needed for tests (the existing tuple) and the extras as well.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,12 +5,12 @@ import os
 
 EXTRA_REQS = ("pandas", "chardet", "openpyxl")
 TEST_REQS = (
-        "pytest>=3.6",
-        "pytest-cov",
-        "coverage",
-        "codecov",
-        "pytest-benchmark",
-        "black",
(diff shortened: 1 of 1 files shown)


dcslagel (changes requested):
I don't think the type casting (to 'list' and then to 'tuple') is needed.  At least for python 3.10.  In my testing the following  created tuple2 as a tuple2 with all elements in the new tuple:

```
tuple1 = ('other', 'another')  
tuple2 = ('this', 'that') + tuple1
type(tuple2)
<class 'tuple'>  
print(tuple2)
('this', 'that', 'other', 'another') 
```  
 
Did you find a reason that the type casting is needed?

kinverarity1:
Thanks! I'm stuck in the past, didn't realise this was no longer a problem. I'll revise the PR.

PR #573, opened 2023-05-17
Update docs with correct supported versions

connortann:
Small PR to update the PyPI classifiers and docs to show support for python 3.11.

On a related note, if you were to adopt the [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html) recommended deprecation policy, it is now time to drop support for 3.7 and 3.8. I didn't do that as part of this PR though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 
 Read and write Log ASCII Standard files with Python.
 
-This is a Python 3.5+ package to read and write Log ASCII Standard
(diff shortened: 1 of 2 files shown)


dcslagel (approved):

dcslagel:
Issue https://github.com/kinverarity1/lasio/issues/574 is created to track End Lasio support for Python 3.7.

kinverarity1:
Thanks @connortann and @dcslagel

PR #579, opened 2023-09-07
30% faster reads by removing unnecessary `file.tell()` calls

emilykl:
Rather than calling `file.tell()` for every line while searching for section headers, call it only for lines containing a section header. This results in a _significant_ read performance boost of at least 30%.

| [Benchmark before](https://github.com/kinverarity1/lasio/actions/runs/5914132966/job/16039111154)  | [Benchmark after](https://github.com/emilykl/lasio/actions/runs/6113355365/job/16592688444) | Improvement |
| ------------- | ------------- | ------------- |
| 264ms  | 149ms  | 43% |

(This particular benchmark shows an even greater improvement but I was usually seeing values closer to 30% during local testing.)

This change is less straightforward than it seems, because we need to know the file position of the beginning of the header line; but calling `file.tell()` after reading that line gives us the file position of the beginning of the _following_ line.

This approach reverse-engineers the correct file position by observing that `file.tell()` seems to correspond to  subtracting the length of the current line, plus an adjustment depending on the file's line ending type (CRLF vs. LF).

Since the performance improvement is so large, I think this change is worth pursuing. However, there are two caveats:

- This approach _does not_ work reliably on files with mixed line endings.
  - As far as I can tell, Python provides no way of finding out the line ending _of a specific line_. `file_obj.newlines` only reveals which line endings have been encountered so far in the file.
  - [`test_open_url_different_newlines()`](https://github.com/emilykl/lasio/blob/41dc97316ae88a0dcff5c1f90c5bfcb389c7c920/tests/test_open_file.py#L37C22-L37C22) _happens_ to work because the header line has the same line ending as the first line in the file. Header lines which have a different line ending from the first line in the file will fail.

- And the more serious issue... `io.TextIOWrapper.tell()` officially returns "[an opaque number](https://docs.python.org/3/library/io.html#io.TextIOWrapper.tell)" which is guaranteed only to work as input to `io.TextIOWrapper.seek()`. So even though it seems to correspond to character count in practice, that's not officially documented and could change, making this a fragile approach.

Very open to comments and ideas! If there is a safer way to get the correct file position, without calling `file.tell()` for every line, I would love to implement it.

### DIFF
--- a/lasio/reader.py
+++ b/lasio/reader.py
@@ -297,7 +297,7 @@ def find_sections_in_file(file_obj):
         or line breaks.
 
     """
-    file_pos = int(file_obj.tell())
+    file_pos = 0
     starts = []
     ends = []
     line_no = -1
@@ -314,9 +314,10 @@ def find_sections_in_file(file_obj):
                 ends.append(line_no - 1)
 
     ends.append(line_no)
-    section_positions = []
-    for j, (file_pos, first_line_no, sline) in enumerate(starts):
-        section_positions.append((file_pos, first_line_no, ends[j], sline))
+    section_positions = [
+        (file_pos, first_line_no, ends[j], sline)
+        for j, (file_pos, first_line_no, sline) in enumerate(starts)
+    ]
     return section_positions
 
 
@@ -329,7 +330,11 @@ def full_line_length(line, file_obj):
         else file_obj.newlines[0]
     )
     newline_adjust = len(newlines) - 1 if newlines is not None else 0
-    return len(line) + newline_adjust
+    if line.endswith("\n"):
+        length = len(line) + newline_adjust
+    else:
+        length = len(line)
+    return length
 
 
 def determine_section_type(section_title):
@@ -602,6 +607,7 @@ def read_data_section_iterative_numpy_engine(file_obj, line_nos):
         names=None,
         unpack=True,
         loose=False,
+        dtype=float,
     )
 
     # If there is only one data row, np.genfromtxt treats it as one array of
--- a/lasio/reader.py
+++ b/lasio/reader.py
@@ -300,24 +300,36 @@ def find_sections_in_file(file_obj):
     file_pos = int(file_obj.tell())
     starts = []
     ends = []
-    line_no = 0
-    line = file_obj.readline()
-    # for i, line in enumerate(file_obj):
+    line_no = -1
+    line = True
+
     while line:
+        line = file_obj.readline()
+        line_no = line_no + 1
         sline = line.strip().strip("\n")
         if sline.startswith("~"):
+            file_pos = int(file_obj.tell()) - full_line_length(line, file_obj)
             starts.append((file_pos, line_no, sline))
             if len(starts) > 1:
                 ends.append(line_no - 1)
-        file_pos = int(file_obj.tell())
-        line = file_obj.readline()
-        line_no = line_no + 1
 
     ends.append(line_no)
     section_positions = []
     for j, (file_pos, first_line_no, sline) in enumerate(starts):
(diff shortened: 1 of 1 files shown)


dcslagel:
@emilykl @kinverarity1 ,  

This seems like a really good idea.  It might be worthwhile to revisit how mixed-line endings are currently handled.  If they could be standardized on `open()` (or some other way) then the by the time the code gets to `full_line_length(line, file_obj)` the line-endings would be reliably one type and this pull-request would be a great addition.

For the issue around io.TextIOWrapper.tell() being an opaque number for io.TextIOWrapper.seek(),  it feels like we could accept this risk and either: hope new versions of Python will always handle it in some proper way of, or update Lasio code at such a time that Python changes the io.TextIOWrapper.tell()/seek() behavior.  Does that seem like the right approach to you?

kinverarity1:
@emilykl Thank you! This is wonderful.  I'm happy to merge if we can work out a way to provide for a sensible exception and a backup way to read files with mixed line endings (surely they are uncommon) - such as a keyword argument.

@dcslagel I'm on board with all your ideas. What do you think, Should we all get this PR to a merge state and release it first while considering other improvements for later work?

30% is worth risking changes with python in the future in my opinion.

emilykl:
Thanks for the feedback! Very appreciated.

> If they could be standardized on open() (or some other way) then the by the time the code gets to full_line_length(line, file_obj) the line-endings would be reliably one type and this pull-request would be a great addition.

What we could potentially do is _disable_ newline standardization by [passing the argument `newline=''` to `open()`](https://docs.python.org/3/library/io.html#io.TextIOWrapper). That would make `readline()` return the exact line endings for each line as written on disk, rather than standardizing, so we wouldn't have to guess.

The only issue I see there is that the `LASfile()` constructor accepts _either_ a file path or an already-opened file object, so if the user has passed an already-opened file we won't know whether the newlines are being standardized or not, so it may fail completely in that case.

In any case, I think it probably makes sense to make this "fast" method a configurable option -- either opt-in or opt-out. We can document the caveats, and perhaps even add a hard restriction that if the "fast" method is enabled, `file_ref` MUST be a path, not an open file object.

kinverarity1:
Sounds good!

> In any case, I think it probably makes sense to make this "fast" method a configurable option -- either opt-in or opt-out. We can document the caveats, and perhaps even add a hard restriction that if the "fast" method is enabled, file_ref MUST be a path, not an open file object.

I'd prefer the "fast" option to be on by default. 

Also perhaps it makes sense to remove the support for file_ref to be an open file object at the same time. If I understand correctly, this method should still work then in the two scenarios of file_ref being a path or a string containing the LAS file contents, correct?

dcslagel:
@kinverarity1, @emilykl 

> What do you think, Should we all get this PR to a merge state and release it first while considering other improvements for later work?
> 30% is worth risking changes with python in the future in my opinion.

------------------------------------------------------

1. We should probably make the failing case for 

> [test_open_url_different_newlines()](https://github.com/emilykl/lasio/blob/41dc97316ae88a0dcff5c1f90c5bfcb389c7c920/tests/test_open_file.py#L37C22-L37C22) happens to pass anyway because the header line has the same line ending as the first line in the file. Header lines which have a different line ending from the first line in the file will fail.   

So a file that has a header line with a different line-ending from the first line in the file, plus a test case for it.  So that it is a visible issue.

2. Currently,  I don't think we should remove the support for file_ref to be an open file object.  It may be a popular feature.  If we do want to consider removing support for file_ref, it is probably worth asking the user base.

3. It looks like the new function `full_line_length(line, file_obj):`. could be expanded to handle the main variations of line endings.  Is that reasonable or is it difficult?

PR #581, opened 2023-11-17
added version 1 to ORDER_DEFINITIONS

jwildman135:
- Version 1 LAS files have the same ORDER_DEFINITIONs as Version 1.2 LAS files.
- Currently Version 1 LAS files produce a key error.
- I propose adding Version 1 LAS files to the ORDER_DEFINITIONS, with the same definition as Version 1.2 LAS files, to allow support for this version.

### DIFF
--- a/lasio/defaults.py
+++ b/lasio/defaults.py
@@ -44,6 +44,32 @@ def get_default_items():
 
 
 ORDER_DEFINITIONS = {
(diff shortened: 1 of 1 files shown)


kinverarity1 (approved):
Thank you very much for this PR. I apologise for the delay in approving.

PR #596, opened 2024-05-14
Remove print statement left in stack_curves method

ltbrizolara:
A print statement was left in stack_curves method, printing to the console the list of all mnemonics to be stacked. And they can be thousands...


kinverarity1 (approved):
Thank you! :heart:

PR #600, opened 2025-02-08
Upgrade imports to supported python versions

frank1010111:
Python 3.9 is the lowest supported Python version. In this code, I've changed the imports to follow ruff's isort rules and remove the dance required for supporting Python 2 and Python 3 less than 3.9.

The unit tests pass on my local linux machine.

Hope this helps.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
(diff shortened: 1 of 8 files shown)


ThomasMGeo:
Also see no reason why this should not be bumped to 3.10, the latest versions of numpy do not support 3.9.

kinverarity1 (approved):

kinverarity1:
Thank you very much @frank1010111 and apologies for the delay in merging.

PR #602, opened 2025-03-19
Fix: Correct file reading logic for single row/column detection

gvaleras:
The check array.shape == 1 cannot distinguish between a single row and a single column. Using if max_rows == 1 correctly resolves this ambiguity, ensuring the array is reshaped as expected.

### DIFF
--- a/lasio/reader.py
+++ b/lasio/reader.py
@@ -567,13 +567,15 @@ def read_data_section_iterative_numpy_engine(file_obj, line_nos):
         file_obj, skip_header=first_line, max_rows=max_rows, names=None, unpack=True, loose=False
     )
 
-    # If there is only one data row, np.genfromtxt treats it as one array of
-    # individual values. Lasio needs a array of arrays. This if statement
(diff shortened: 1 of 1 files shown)


gvaleras:
Hello @kinverarity1, could you please review this PR when possible? It needs to be rebased with the latest main, and I noticed that python3.7 (which caused the CI failure) and python3.8 have been dropped from main since then. Additionally, could you please approve the workflow so it can run? Thank you!

kinverarity1:
Sorry the review button has disappeared. Could you change the style to avoid in-line if?

```

kinverarity1:
Thank you for working on it

gvaleras:
Hello @kinverarity1, I have rebased the branch with main and applied your requested changes, replacing the inline if with a block if-else. Please have a look and let me know if anything else is needed. Thank you!

kinverarity1 (approved):

kinverarity1:
Thank you @gvaleras

